PR #6, opened 2024-02-25
Update WAKU2-METADATA: Move to draft

jimstir:
Move 66/WAKU2-METADATA to draft.

### DIFF
--- a/waku/standards/core/66/metadata.md
+++ b/waku/standards/core/66/metadata.md
@@ -1,0 +1,37 @@
+---
+title: WAKU2-METADATA
+name: Waku Metadata Protocol
+editor: Alvaro Revuelta <alrevuelta@status.im>
+contributors:
+---
+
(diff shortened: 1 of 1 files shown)


jm-clius (approved):
Makes sense to me. Should we continue maintaining the legacy number here for new draft protocols? No strong opinion on my side, but whereas for filter, store, lightpush, etc. the numbers have become idiomatic, I'm less sure it's the case for some of these later additions.

jimstir:
> Makes sense to me. Should we continue maintaining the legacy number here for new draft protocols? No strong opinion on my side, but whereas for filter, store, lightpush, etc. the numbers have become idiomatic, I'm less sure it's the case for some of these later additions.

I decided to use the same numbers for the three specs you mentioned should become draft specs before the repo transition. Following specs that migrate from waku/specs will be given new numbers.

PR #16, opened 2024-02-29
wip: use noise instead of ECIES

jimstir:
Ref: https://github.com/waku-org/examples.waku.org/issues/240

Reference pull request: https://github.com/vacp2p/rfc/pull/606


fryorcraken:
Can be closed for now. I think we need to aim for a more general solution/API on Waku side

PR #38, opened 2024-05-14
Updating COSS: Stagnant Status

jimstir:
Remove delete status and retired status. Since specifications will remained archived, using stagnant status is more neutral status to refer to deleted and retired specifications. Also stagnant specification can be awarded previous status if deleted status was premature.

### DIFF
--- a/vac/1/coss.md
+++ b/vac/1/coss.md
@@ -131,18 +131,15 @@ When stable specifications are replaced by newer draft specifications, they beco
 Deprecated specifications should not be changed except to indicate their replacements, if any.
 Deprecated specifications are contracts between editors, implementers and end-users.
 
-### Retired Specifications
+### Stagnant Specifications
 
-When deprecated specifications are no longer used in products, they become **retired** specifications.
-Retired specifications are part of the historical record.
-They should not be changed except to indicate their replacements, if any.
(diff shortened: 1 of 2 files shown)


jimstir:
> @jimstir please, have all specs from https://specs.status.im been migrated yet? And could you create an issue to removed references to https://specs.status.im for those already moved here?

@felicio 👍 I'll get on this, https://github.com/vacp2p/rfc-index/issues/96

jimstir:
> I think this change makes sense. I can't really see the value of a `retired` status on top of `deprecated`, but we certainly have many `stagnant` specs. My only question is who would manage and archive the `stagnant` RFCs, Vac or the individual projects, such as Waku? Would `raw` specs that are still managed by the subprojects be `stagnant` within those same project repos or be identified and moved to Vac indexing to be archived here?

If project repo decide to use the `stagnant` status, they should archive that spec within the project repo after long periods of no updates (or when agreement occurs from discussion). The RFC team can assist the management of archiving specs(getting approval from listed contributors) if the project team request it. RFC-Index does host some raw specs (Vac, Status, and in the future Nomos) so the same will apply to those specs that live here. @jm-clius

Cofson:
Can we merge this one @jimstir ?

PR #42, opened 2024-05-29
Update 13/STORE: new error codes for Extended HistoryResponse

NagyZoltanPeter:
This is a small adjustment on new error codes introduced for waku's store protocol response.
Change is introduced by https://github.com/waku-org/pm/issues/66 implementation.

### DIFF
--- a/waku/standards/core/13/store.md
+++ b/waku/standards/core/13/store.md
@@ -105,6 +105,8 @@ message HistoryResponse {
   enum Error {
     NONE = 0;
     INVALID_CURSOR = 1;
+    TOO_MANY_REQUESTS = 429;
+    SERVICE_UNAVAILABLE = 503;
   }
   Error error = 4;
 }


jimstir (approved):

jm-clius (commented):
I apologise for confusion with new and existing RFCs and different repos, but indeed we are replacing this RFC 13 with a new version of Store (v3) where we do not explicitly define the error codes (exactly because we expect the list to grow significantly and do not want to have to update the RFC every time). Store v3 spec however still only exists in a branch as we have been modifying it extensively during the POC implementation. However, I think we are close to merging it into waku.org/specs as a new spec (but not RFC).

TL;DR - we not need to update this spec as we're moving to Store v3: https://github.com/vacp2p/rfc/pull/665/files

jimstir:
> I apologise for confusion with new and existing RFCs and different repos, but indeed we are replacing this RFC 13 with a new version of Store (v3) where we do not explicitly define the error codes (exactly because we expect the list to grow significantly and do not want to have to update the RFC every time). Store v3 spec however still only exists in a branch as we have been modifying it extensively during the POC implementation. However, I think we are close to merging it into waku.org/specs as a new spec (but not RFC).
> 
> TL;DR - we not need to update this spec as we're moving to Store v3: https://github.com/vacp2p/rfc/pull/665/files

@NagyZoltanPeter I will close this pr, when Store v3 is ready it will live in waku/specs repo.

NagyZoltanPeter:
> > I apologise for confusion with new and existing RFCs and different repos, but indeed we are replacing this RFC 13 with a new version of Store (v3) where we do not explicitly define the error codes (exactly because we expect the list to grow significantly and do not want to have to update the RFC every time). Store v3 spec however still only exists in a branch as we have been modifying it extensively during the POC implementation. However, I think we are close to merging it into waku.org/specs as a new spec (but not RFC).
> > TL;DR - we not need to update this spec as we're moving to Store v3: https://github.com/vacp2p/rfc/pull/665/files
> 
> @NagyZoltanPeter I will close this pr, when Store v3 is ready it will live in waku/specs repo.

Yes, its ok! Thanks for the heads up.

PR #85, opened 2024-07-30
Mix Protocol Spec

AkshayaMani:


vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+
+In the original
+[Sphinx](https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf) paper, the
+authors use node IDs of size $\kappa$ ($16$ bytes) to represent the next hop
+addresses. To accommodate larger addresses, we'll use a combined size of
+$t\kappa$ bytes for the address and delay, where $t$ is small (_e.g.,_ $t = 2$
+or $3$).
+
+- **Delay**: 2 bytes
+  Allows delays up to 65,535 milliseconds ≈ 65 seconds.
```
rymnc, line 270: This is quite interesting, I initially thought RLN could be used for rate limiting in this context but due to the random delays that may be introduced at each intermediary node, verifying RLN proofs becomes difficult since they are strictly coupled to the sender timestamp

rymnc: (in the context of waku)

vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+routing information.
+
+```proto
+message SphinxPacket {
+  bytes alpha = 1; // 32 bytes
+  bytes beta = 2; // 192 - 208 bytes
+  bytes gamma = 3; // 16 bytes
+  bytes delta = 4; // variable size, max 2157 bytes
+}
+```
```
rymnc, line 296: 👍

vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+   i. **Create Challenge**
+
+   Retrieve the current Unix timestamp, `timestamp`, in seconds (4 bytes).
+
+   ii. **Find A Valid Nonce**
+   - Initialize the `nonce` to a 4-byte value (initially zero).
+   - Increment the `nonce` until the SHA-256 hash of
+    `<libp2p_message_bytes | timestamp | nonce>` has at least
+    18 leading zeros.
+   - The final value of the `nonce` is considered valid.
```
rymnc, line 358: since most of the projects within the IFT are targeting resource constrained devices, perhaps we should do an analysis on how much time it takes for cheaper devices ~ but then we come back to the same issue that [whisper](https://github.com/ethereum/whisper) and Status v1 had, i.e if we reduce the difficulty too much, then malicious actors with beefier machines can participate in spamming the network.

vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+
+### normative
+
+[libp2p](https://libp2p.io)\
+[Sphinx](https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf)
+
+### informative
+[PoW](https://bitcoin.org/bitcoin.pdf)\
+[Sphinx packet size](https://petsymposium.org/popets/2024/popets-2024-0050.pdf)
+
```
rymnc, line 658:
```suggestion
### normative

- [libp2p](https://libp2p.io)\
- [Sphinx](https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf)

### informative
- [PoW](https://bitcoin.org/bitcoin.pdf)\
- [Sphinx packet size](https://petsymposium.org/popets/2024/popets-2024-0050.pdf)

```

rymnc (commented):
Content looks great to me! An amazing RFC, well written and understandable from implementer PoV. 

Some questions and notes in-line 👍

vac/raw/mix.md
AkshayaMani: The reasoning for 18 leading zeroes is mentioned in the Notion page. I can integrate it in here:

> Difficulty Level: The difficulty level is usually expressed as the number of leading zeros required in the hash output. It is chosen such that the computational effort required is significant but not prohibitive. We recommend a reasonable difficulty level that requires around 16-18 leading zeros in the SHA-256 hash. This would roughly take 0.65 to 2.62 seconds to compute in a low-grade CPU, capable of computing 100,000 hashes per second.

vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+  the list of discovered nodes.
+
+- **Forwarding To Next Hop (Intermediary Nodes Only)**
+
+  When a mix node receives an incoming Sphinx packet, it performs the following
+  actions:
+
+  i. Decrypts the packet to obtain the next hop multiaddress\
+  ii. Checks if the next hop is in the list of discovered nodes.\
+  iii. If not, performs discovery for that specific node.\
```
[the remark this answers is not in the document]
AkshayaMani: Great question! The intermediary node cannot choose any arbitrary node because the Sphinx packets are encrypted to specific nodes, and any arbitrary node would not be able to decrypt them. Notifying the sender through secure backtracking is also not straightforward, as the failure message would need to be encrypted and sent back through the established mix route in reverse. However, each intermediary node only knows the previous node, not the entire path back to the sender.

An easier solution is for the sender to send messages through multiple paths (~3 to 4) when reliability is a concern. In fact, for the Anonymized GossipSub Protocol (AGP) [WiP], it is recommended that the sender chooses 3-4 message paths. I'll add this to the spec.

vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+It achieves this by using the Sphinx packet format, which encrypts and routes
+messages through a series of nodes (mix nodes) before reaching the recipient.
+
+Key features of the protocol include:
+
+i. Path selection for choosing a random route through the network via multiple
+mix nodes.\
+ii. Sphinx packet construction and processing, providing cryptographic
+guarantees of anonymity and security.\
+iii. Proof of Work mechanism to prevent spamming of the mix network.\
```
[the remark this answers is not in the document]
AkshayaMani: Yes, we can use a different method of spam prevention, such as RLN. However, given the need for a decentralized, simple, and independent spam prevention mechanism in our custom mix protocol, PoW is the preferred choice.

i. PoW does not require a trusted setup or continuous management. It operates independently without relying on any central or semi-trusted authority. 
ii. It does not have dependencies on external systems or blockchains, avoiding the overhead and potential points of failure associated with such integrations. This ensures that our mix protocol remains self-contained and robust.
iii. It provides a straightforward security model by requiring computational effort to send messages. It is simple to implement and understand, making it easier to integrate into our mix protocol without introducing additional complexity.
iv. When implemented properly, PoW scales well. The exit nodes can efficiently verify a large number of incoming messages without significant performance degradation.

vac/raw/mix.md
```diff
@@ -0,0 +1,658 @@
...
+
+#### 2.3 Incentives
+
+To publish an anonymous libp2p message (_e.g.,_ GossipSub, Ping, etc.), nodes
+must run a mix node instance. This requirement serves as an incentive for nodes
+to participate in the mix network, as it allows them to benefit from the
+anonymity features while also contributing to the network's overall anonymity
+and robustness.
+
+#### 2.4 Node Discovery
```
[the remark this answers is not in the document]
AkshayaMani: Yes, that would be the best approach. For initial testing and PoC deployments, we could use a static list of mix nodes. This would simplify the overall setup and allow for controlled experiments to validate the mix protocol's core functionality, streamlining development and testing.

Additionally, not committing to a specific discovery mechanism at this stage provides flexibility for future integrations with various discovery methods, enabling better interoperability. Further, this approach would allow us to design and incorporate a robust node discovery system incrementally as the protocol matures.

It is added in the spec to ensure that any future discovery mechanisms used with the mix protocol i) guarantee that the discovered mix nodes represent a random sample of the entire network and ii) accommodate required mix node discovery information. These are specific needs of the mix protocol and may not be applicable to other libp2p protocols.

vac/raw/mix.md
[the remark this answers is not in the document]
rymnc: totally understand now, but in the event that there is a malicious actor with large compute capacity, will the network be able to handle the load?
This is what killed whisper

vac/raw/mix.md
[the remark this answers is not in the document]
rymnc: agree, thanks for the detailed explanation!

vac/raw/mix.md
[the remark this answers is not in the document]
rymnc: Sounds good, this works, maybe we should have a heuristic weighing the number of paths vs network overhead and reliability of message delivery

vac/raw/mix.md
rymnc: Nice, thank you

vac/raw/mix.md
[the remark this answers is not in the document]
AkshayaMani: Agreed! A heuristic that balances the number of paths against network overhead and message delivery reliability is valuable. This can be evaluated through simulations to determine if, for example, using four paths provides sufficient reliability while managing overhead effectively.

However, this aspect could be omitted from the initial implementation. It is something that top-level applications can configure based on their specific requirements for anonymity, reliability, and performance trade-offs. This would allows us to focus on core functionality while providing flexibility for application-specific tuning later on.

vac/raw/mix.md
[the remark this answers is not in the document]
AkshayaMani: This is a valid question! Our Mix Protocol improves on Whisper by centralizing PoW verification to exit nodes rather than every node along the message path. This reduces the computational load to a linear relationship between the number of messages and verifications, rather than the multiplicative burden seen in Whisper.

While this design helps mitigate computational strain, exit nodes are still vulnerable to targeted attacks. To counter this, PoW difficulty can be dynamically adjusted based on network load, with updated levels communicated via ENRs. Senders can then perform discovery to obtain the current ENR of the chosen exit node(s), ensuring they use the appropriate PoW difficulty. This would help the network adapt to varying conditions. This approach enhances resistance to spam, though it is worth noting that similar vulnerabilities are faced by other anonymity networks as well.

Despite these limitations, PoW is a practical choice due to its decentralized, simple, and independent nature. While alternative mechanisms might offer more robustness, they require trust or blockchain dependencies. Thus, PoW offers a pragmatic solution for initial deployment and can be refined over time to provide more resistance against high-compute capacity attacks.

rymnc (approved):
LGTM, cc: @kaiserd for final review

vac/raw/mix.md
```diff
@@ -0,0 +1,659 @@
+---
+title: MIX
+name: Mix
```
jimstir, line 3: @AkshayaMani The title should be `MIX-PROTOCOL` and name `The Mix Protocol`.

AkshayaMani: Done!

vac/raw/mix.md
```diff
@@ -0,0 +1,659 @@
...
+- It can be an intermediary node for messages it is forwarding.
+- It can be an exit node for messages it is disseminating.
+
+#### 2.3 Incentives
+
+To publish an anonymous libp2p message (_e.g.,_ GossipSub, Ping, etc.), nodes
+MUST run a mix node instance. This requirement serves as an incentive for nodes
+to participate in the mix network, as it allows them to benefit from the
+anonymity features while also contributing to the network's overall anonymity
+and robustness.
```
seugu, line 117: In case there is no enough mixnodes in a network, may it cause some anonymity leakages for mixnodes? If so, is there any theroetical threshold percentage for being safe?

vac/raw/mix.md
```diff
@@ -0,0 +1,659 @@
...
+### 3. Cryptographic Primitives and Security Parameter
+
+- **Security Parameter:** $\kappa = 128$ bits provides a balance between
+  security and efficiency.
+- **Cryptographic Primitives**
+  - **Group G**: Curve25519 elliptic curve offers 128-bit security with small
+    (32-byte) group elements, efficient for both encryption and key exchange.
+  - **Hash function H**: SHA-256.
+  - **KDF:** SHA-256 (truncated to 128 bits).
+  - **AES-CTR:** AES-128 in counter mode.
```
seugu, line 214: Is there any specific reason to use CTR mode?

vac/raw/mix.md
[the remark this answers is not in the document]
AkshayaMani: Using Verifiable Delay Functions (VDFs) as an alternative to Proof of Work (PoW) is an interesting idea. VDFs require sequential computation, which eliminates the parallel computing advantages of users with high computational power, creating a more level playing field. They also provide efficient verification and deterministic output, which can enhance the protocol's performance.
However, implementing VDFs may introduce additional complexity and could require a trusted setup, raising new security considerations. While VDFs are promising for future versions of the Mix protocol, we should carefully evaluate their maturity and practical applications before deciding to integrate them.

vac/raw/mix.md
AkshayaMani: You raise an important point. Indeed, having too few mix nodes can potentially lead to anonymity leakages, and this is not exclusive to our mix protocol but a general concern for all mix networks. While there's no universally agreed-upon threshold, the general principle is that more mix nodes provide better anonymity. 

To mitigate risks in smaller networks, we could consider (in the later versions):
1. Implementing stricter node selection criteria
2. Introducing additional cover traffic

We should keep this in mind as we develop and deploy the protocol.

AkshayaMani:
New PR can be found here: [libp2p Mix Protocol Spec Draft #97](https://github.com/vacp2p/rfc-index/pull/97)

PR #87, opened 2024-08-14
feat: status mvds usage

kaichaosun:
Replace this PR: https://github.com/waku-org/specs/pull/31

Fix: https://github.com/status-im/status-go/issues/5694

### DIFF
--- a/status/raw/status-mvds.md
+++ b/status/raw/status-mvds.md
@@ -19,6 +19,12 @@ Status app uses MVDS to ensure messages going through Waku are acknolwedged by t
 
 ## Message types
 
+Various Message Types contain distinct information defined by the app to facilitate convenient serialization and deserialization.
+
+E2E reliability is a feature that ensures messages are delivered to the recipient. This is initially achieved by using MVDS in Status.
+
+Chat Type specifies the category of chat that a message belongs to. It can be OneToOne (aka Direct Message), GroupChat, or CommunityChat. These are the three main types of chats in Status.
+
 | Message Type                                                               | Use MVDS                            | Need e2e reliability | Chat Type               |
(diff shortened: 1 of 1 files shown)


jm-clius (commented):
Thanks! This has already been a very useful reference list for me. :)

I'm not sure though that a reference list like this should be a specification/RFC on its own. Generally a specification (even a Best Current Practice), serves to introduce a context, explain terms and either recommend or mandate how these should be applied for certain use cases. For example, concepts such as "e2e reliability" or "chat type" is not defined in this spec and the relevance is not clear. This list will also likely change quite often as we continue working on reliability, add new Status protocols, etc.

For me, the direction in which Status specs should be going, could be:
1. a spec for each reliability mechanisms existing within Status context
2. a spec for the encryption paradigms within Status context
3. separate specifications for Status App procedures (i.e. 1:1, Communities, Push Notifications, Sync, etc. should not be mixed together) where the procedures are defined as well as the reliability mechanisms at play, with reference to the reliability and encryption specs.

Of course, all of this is a large undertaking and out of scope now. I'm happy to approve if we believe there's value in temporarily keeping this reference list as a spec on its own. I might not be up to date with current plans for what we want to spec in terms of reliability and don't want to block progress (cc @fryorcraken @plopezlpz)

kaichaosun:
@jm-clius added the missing context information.

IMO this spec works like a lookup table when we deal with how Status is using the reliability solution (MVDS here). It still not find its main focus but still valuable for not losing like keep in google sheet and keep improving the table.

Thanks for pointing out the missing specs, there probably are some content already lives somewhere, but we need to put them together and reference each other with updated links.

jimstir (approved):

kaichaosun:
@jm-clius is there any blocker for this to be merged?

jm-clius (approved):

PR #114, opened 2024-12-06
Update peer-exchange.md to fix a build error

jinhojang6:
Screenshot 2024-12-06 at 8 29 20 PM

Images should be defined using relative paths to successfully build https://rfc.vac.dev/, which is powered by Docusaurus.

I have verified that this update resolves the build issue locally.


jimstir (approved):
Approved, also, this spec does not show on the rfc.vac.dev website.

PR #118, opened 2024-12-17
update communities spec as per content topic changes

chaitanyaprem:
- [x] Update content topic usage as per https://forum.vac.dev/t/status-communities-review-and-proposed-usage-of-waku-content-topics/335
- [x] Update symmetric encryption done at content topic level before messages are handed over to waku
- [x] Update about shard or pubsub topic usage
~- [ ] update communities flows as per current understanding~

Anything else that can be added please let me know.

@jm-clius  i have tried to brief about community shards in this spec, not sure if that in itself be a separate spec with a ref included here. Please advise.

### DIFF
--- a/status/56/communities.md
+++ b/status/56/communities.md
@@ -423,7 +423,7 @@ contentTopic = "/waku/1/0x" + topic + "/rfc26"
 
 #### Community channels/chats
 
-All channels/chats shall use a single content-topic which is derived based on a universal chat id ir-respective of their individual unique chat ids.
+All channels/chats shall use a single content-topic which is derived from a universal chat id irrespective of their individual unique chat ids.
 
 ### Community Management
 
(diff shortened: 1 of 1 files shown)


status/56/communities.md
```diff
@@ -379,6 +401,30 @@ for i = 0; i < topicLen; i++ {
...
+for i = 0; i < topicLen; i++ {
+    topic[i] = hash[i]
+}
+
+contentTopic = "/waku/1/0x" + topic + "/rfc26"
+```
+
+#### Community channels/chats
+
+All channels/chats shall use a single content-topic which is derived based on a universal chat id ir-respective of their individual unique chat ids.
```
jm-clius, line 426:
```suggestion
All channels/chats shall use a single content-topic which is derived from a universal chat id irrespective of their individual unique chat ids.
```

jm-clius (approved):
While I think that we may need to do a full revision of the spec in future, this makes sense to me in the interim to specify content topic usage. Thanks!

PR #121, opened 2025-01-06
rearranging the org within Nomos folder

Cofson:


jimstir:
@Cofson I keep forgetting, the number should be removed, this spec never became draft.

jimstir (approved):

PR #122, opened 2025-01-21
update secpm rfcs

seugu:
Updated the RFCs by removing three out of four by keeping [eth-demls](https://github.com/vacp2p/rfc-index/blob/main/vac/raw/eth-demls.md) the rest of the three are kept in the [backup repo](https://github.com/vacp2p/rfc-index/tree/mls-backup).


seugu:
Cancelled because we re-organize this RFC.

PR #125, opened 2025-02-06
10/WAKU2: Update

jimstir:
Updates include changing links, removing deprecated specs, and adding updated specs.

### DIFF
--- a/waku/standards/core/10/waku2.md
+++ b/waku/standards/core/10/waku2.md
@@ -236,7 +236,7 @@ These fields MUST be used under certain conditions, as set out in the specificat
 Both EIP-1459 DNS-based discovery and Node Discovery v5 operate on ENR,
 and it's reasonable to expect even wider utility for ENR in Waku v2 networks in the future.
 
-### Request/Reply Domain
+### Request/Response Domain
 
 In addition to the Gossip domain,
 Waku v2 provides a set of request/response protocols.
--- a/waku/standards/core/10/waku2.md
+++ b/waku/standards/core/10/waku2.md
@@ -395,14 +395,14 @@ the actors / actors' PII. Concerning anonymity, Waku v2 provides the following f
 **Publisher-Message Unlinkability**:
 This feature signifies the unlinkability of a publisher
 to its published messages in the 11/WAKU2-RELAY protocol.
-The [Publisher-Message Unlinkability](../11/relay.md/#security-analysis)
+The [Publisher-Message Unlinkability](waku/standards/core/11/relay.md/#security-analysis)
 is enforced through the `StrictNoSign` policy due to which the data fields
 of pubsub messages that count as PII for the publisher must be left unspecified.
 
 **Subscriber-Topic Unlinkability**:
 This feature stands for the unlinkability of the subscriber
 to its subscribed topics in the 11/WAKU2-RELAY protocol.
-The [Subscriber-Topic Unlinkability](../11/relay.md/#security-analysis)
+The [Subscriber-Topic Unlinkability](waku/standards/core/11/relay.md/#security-analysis)
 is achieved through the utilization of a single PubSub topic.
 As such, subscribers are not re-identifiable from their subscribed topic IDs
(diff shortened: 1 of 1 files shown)


waku/standards/core/10/waku2.md
```diff
@@ -14,37 +14,37 @@ contributors:
 
 ## Abstract
 
-Waku v2 is family of modular peer-to-peer protocols for secure communication.
+Waku v2 is a family of modular peer-to-peer protocols for secure communication.
```
jm-clius, line 17: Mmm, I realise now I forgot to comment about this while reviewing the WAKU2-MESSAGE spec, but we've generally stopped referring to "Waku v2" and simply use "Waku". The "v2" seems to confuse people, as v1 (now usually just "legacy Waku") is implemented nowhere, difficult to find specifications for and not actively developed. We do want to acknowledge though that this spec introduced a newly designed Waku stack. Perhaps it's possible to drop the "Waku v2" but add a single comment at the top to explain that this suite of protocols were previously referred to as "Waku v2" in order to distinguish it from the obsolete legacy Waku protocol stack?

waku/standards/core/10/waku2.md
```diff
@@ -117,19 +123,19 @@ Since these aren't negotiated libp2p protocols,
...
 
 There are also more experimental libp2p protocols such as:
 
-1. `/vac/waku/swap/2.0.0-beta1`
-2. `/vac/waku/waku-rln-relay/2.0.0-alpha1`
+1. `/vac/waku/waku-rln-relay/2.0.0-alpha1`
+2. `/vac/waku/peer-exchange/2.0.0-alpha1`
 
-These protocols and their semantics are elaborated on in their own specs.
+The semnatics of these protocols are referred to by RFC ID [17/WAKU2-RLN-RELAY](waku/standards/core/17/rln-relay.md) and [34/WAKU2-PEER-EXCHANGE](waku/standards/core/34/peer-exchange.md).
```
jm-clius, line 136:
```suggestion
The semantics of these protocols are referred to by RFC ID [17/WAKU2-RLN-RELAY](waku/standards/core/17/rln-relay.md) and [34/WAKU2-PEER-EXCHANGE](waku/standards/core/34/peer-exchange.md).
```

waku/standards/core/10/waku2.md
```diff
@@ -229,50 +232,50 @@ describes the usage of [EIP-778 ENR (Ethereum Node Records)](https://eips.ethere
...
 `waku2`, that a Waku v2 node MAY use for discovery purposes.
-These fields MUST be used under certain conditions, as set out in the spec.
-Both EIP-1459 DNS-based discovery and Node Discovery v5 operates on ENR,
-and it's reasonable to expect even wider utility for ENR in Waku v2 networks in future.
+These fields MUST be used under certain conditions, as set out in the specification.
+Both EIP-1459 DNS-based discovery and Node Discovery v5 operate on ENR,
+and it's reasonable to expect even wider utility for ENR in Waku v2 networks in the future.
 
-### Request/Reply domain
+### Request/Reply Domain
```
jm-clius, line 239: I think we've since opted for "Request/Response" instead of .../Reply

jm-clius (approved):
Approved, as these changes certainly improve the spec. Main open question is whether (or how) we can stop using the slightly clumsy "Waku v2" everywhere. I also think Research should add this spec to its backlog for a thorough review, as the state of the art looks very different now from when this spec was first compiled. Thanks!

waku/standards/core/10/waku2.md
jimstir: Wouldn't this affect all Waku RFCs and RFC IDs? Also may need consensus from the other contributors named in this spec if that is the case. I think the only conflict may be with the stable spec 11/WAKU2-RELAY which only mention Waku v2 in the name within the header (stable specs can not be updated). 
The way 11/WAKU2-RELAY is written may be an alternative approach where Waku is used throughout the spec, but v2 is only mentioned in the name within the header. This would still require an update of other spec but doable. @jm-clius

PR #127, opened 2025-02-14
Update bindings-api.md

Cofson:


waku/standards/core/36/bindings-api.md
```diff
@@ -1637,7 +1636,7 @@ enr and peerID each node found.
 ```
 
 - 1 - The operation failed for any reason.
-`onErrCb` will be executed with the reason the function execution failed.
+`onErrCb` will be executed with the reason the function execution failed
 - 2 - The function is missing the `onOkCb` or `onErrCb` callback
```
jimstir, line 1640: Why remove the periods?

jimstir (commented):
I see you updated the comment about the period for the one line I mentioned. You make the same changes for multiple lines. Check all your changes for period removal as they may not be necessary. 

This spec is in draft, but the format does not follow the RFC template. Update this to include things like abstract, etc...

PR #130, opened 2025-02-19
docs: add optional retrieval hint to causal history in sds

adklempner:


vac/raw/sds.md
```diff
@@ -137,6 +142,8 @@ include this in the `lamport_timestamp` field.
 and include these in an ordered list in the `causal_history` field.
 The number of message IDs to include in the `causal_history` depends on the application.
 We recommend a causal history of two message IDs.
+* the participant MAY include a `retrieval_hint` in the `HistoryEntry`
+for each message ID in the `causal_history` field.
```
jm-clius, line 146: Perhaps just need more context where this field is first introduced:

```suggestion
for each message ID in the `causal_history` field.
This is an application-specific field to facilitate retrieval of messages,
e.g. from high-availability caches.
```

vac/raw/sds.md
```diff
@@ -206,7 +213,8 @@ in the incoming buffer.
 For each message in the incoming buffer:
 
 * the participant MAY attempt to retrieve missing dependencies from the Store node
-(high-availability cache) or other peers.
+(high-availability cache) or other peers
+using the `retrieval_hint` in the `HistoryEntry`.
```
jm-clius, line 217: Subtle difference, but it's important that retrieval can happen without using the `retrieval_hint`.

```suggestion
(high-availability cache) or other peers.
It MAY use the application-specific `retrieval_hint` in the `HistoryEntry` to facilitate retrieval.
```

vac/raw/sds.md
```diff
@@ -69,12 +69,17 @@ Messages MUST adhere to the following meta structure:
...
 
+message HistoryEntry {
+  string message_id = 1; // Unique identifier of the SDS message, as defined in `Message`
+  optional bytes retrieval_hint = 2; // Optional information to help remote parties retrieve this SDS message; For example, A Waku deterministic message hash or routing payload hash
+}
+
 message Message {
   // 1 Reserved for sender/participant id
   string message_id = 2;          // Unique identifier of the message
   string channel_id = 3;          // Identifier of the channel to which the message belongs
```
jm-clius, line 80: While we're here, should we review the field types for the rest of the `Message`? The choice of `string` for `message_id` and `channel_id` was heavily influenced by similar protocols and current Status code, but might not be the most practical from implementation perspective: I can imagine `bytes` perhaps being more reasonable for `message_id` and `int32` for `channel_id`.

Any opinions here? cc @shash256

jm-clius (approved):
Thanks! This looks great. :)

Added some minor non-blocking comments. Perhaps for a follow-up PR, I would appreciate your and @shash256 feedback on question below re general `Message` field types.

vac/raw/sds.md
adklempner: Opened a separate issue for this: https://github.com/vacp2p/rfc-index/issues/137

vac/raw/sds.md
adklempner: Added in latest commit

vac/raw/sds.md
adklempner: Added in latest commit

PR #109, opened 2024-11-04
Status old RFCs - moving to RFC Index

Cofson:

PR #134, opened 2025-03-06
Status deprecated update2

Cofson:
These are from the previous PR: https://github.com/vacp2p/rfc-index/pull/109


jimstir (commented):
Still have a few issues, check all the specs for each comment.  My mistake, the images folder should be in the deprecated folder as you had it before. I just realized only stable specs should be deprecated ( No change for now). When Hanno reviews, the specs that should stay stable will be changed to stable, and other stable specs would be deprecated. The specs that were draft and raw should be put in a deleted folder.

jimstir (approved):
LGTM, any other changes can be made in future iterations 👍

jazzz (approved):
Appreciate being looped in. 

LGTM - Glad the information is being retained.

PR #144, opened 2025-04-05
update waku/informational/23/topics.md

Cofson:
Updates to topics.md in waku/informational/23 folder


jimstir (approved):
Minor changes. I may have to do another round of updates for this spec and you can review my changes. Also, I do not think you can be added as contributor on RFC updates but you can ask Hanno or Daniel if this is correct. Besides that, LGTM

jm-clius (approved):
Thanks!

PR #154, opened 2025-05-03
add example of llm-generated RFC

adklempner:
made with claude 3.7, extended thinking, explanatory voice

Also fed it all the files under `informational` and `standards` from https://github.com/waku-org/specs using claude's github integration

Original prompt:
```
Review the proposal regarding Semaphore described in this page: https://hackmd.io/@brech1/scale-semaphore-proposal

Then review the existing Waku specs

Then,

Using the Consensus-Oriented Specification System as defined here: https://github.com/vacp2p/rfc-index/blob/main/vac/1/coss.md

Write an RFC for how Semaphore can be used to implement rate-limiting functionality instead of, or alongside, the existing RLN strategy.

Ask qualifying questions if necessary
```

### DIFF
--- a/vac/raw/semaphore-rln.md
+++ b/vac/raw/semaphore-rln.md
@@ -1,0 +1,198 @@
+---
+title: WAKU2-SEMAPHORE-RATE-LIMIT
+name: Semaphore-based Rate Limiting for Waku
+status: raw
+category: Standards Track
+tags: [waku/core-protocol]
+editor: [Your Name]
+contributors:
+---
+
+## Abstract
+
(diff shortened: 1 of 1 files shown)


vac/raw/semaphore-rln.md
```diff
@@ -0,0 +1,198 @@
+---
+title: WAKU2-SEMAPHORE-RATE-LIMIT
+name: Semaphore-based Rate Limiting for Waku
+status: raw
+category: Standards Track
+tags: [waku/core-protocol]
+editor: [Your Name]
+contributors:
+---
+
```
jimstir, line 10: @adklempner I am proposing a specification writing technique RFC [here](https://github.com/vacp2p/rfc-index/pull/155). It is still in draft as I need to fix it up. I included a section introducing prompt recommendations. This RFC and prompt you create could be referenced as a good example. If approved, I think the prompt should be added within the RFC, kind of like a note at the beginning and below the prompt, a statement similar to the description you gave.

PR #157, opened 2025-05-09
Created nomos/raw/sdp.md draft

Cofson:
Created sdp.md draft on nomos/raw folder

### DIFF
--- a/nomos/raw/sdp.md
+++ b/nomos/raw/sdp.md
@@ -1,0 +1,223 @@
+---
+title: SDP
+name: Service Declaration Protocol Specification
+status: raw
+category: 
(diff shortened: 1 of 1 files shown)


jimstir (approved):
LGTM. I did leave one comment, but I do not think this change is necessary now as this is a raw. I do encourage the Nomos team to approve another iteration of this spec to follow the Vac RFC Process, which would include the template, following [1/COSS](https://github.com/vacp2p/rfc-index/blob/nomos/sdp-raw/vac/1/coss.md), adding reference, etc.

danielSanchezQ (approved):
👍

PR #170, opened 2025-08-05
docs: re-add sender ID to messages

jm-clius:
Re-added the concept of a participant ID and the corresponding `sender_id` field in each SDS message.

This is useful to filter "self-triggered" messages as described in https://github.com/waku-org/js-waku/pull/2528

However, this will also be a necessary part of the protocol once p2p message exchange is added.

### DIFF
--- a/vac/raw/sds.md
+++ b/vac/raw/sds.md
@@ -167,7 +167,6 @@ Upon receiving a message,
 
 * the participant SHOULD ignore the message if it has a `sender_id` matching its own.
 * the participant MAY deduplicate the message by comparing its `message_id` to previously received message IDs.
-* the participant MAY deduplicate the message by comparing its `message_id` to previously received message IDs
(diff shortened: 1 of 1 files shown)


vac/raw/sds.md
```diff
@@ -157,6 +165,8 @@ of unacknowledged outgoing messages.
 
 Upon receiving a message,
 
+* the participant SHOULD ignore the message if it has a `sender_id` matching its own
```
fryorcraken, line 168:
```suggestion
* the participant SHOULD ignore the message if it has a `sender_id` matching its own.
* the participant MAY deduplicate the message by comparing its `message_id` to previously received message IDs.
```

fryorcraken (approved):
LGTM

fryorcraken:
@jm-clius btw it's implemented in js-waku :) https://github.com/waku-org/js-waku/pull/2539

PR #171, opened 2025-08-06
Removing `version` usage

fryorcraken:
`version` forces a developer to leak application concerns (encryption used on payload) to the routing layer.

This is unnecessary, as the content topic should be used to discriminate what messages to decrypt with what scheme and key (as it is application dependent).


fryorcraken:
@jimstir @jm-clius @kaiserd can someone remind me how we change a `stable` spec? :D

jimstir:
> @jimstir @jm-clius @kaiserd can someone remind me how we change a `stable` spec? :D

Stable RFC content can not be updated, only the reference links can be changed. You can create a new version of the spec that would go through the RFC process again to become `stable`, once `stable` the current `stable` RFC would become `deprecated`.

As for your update, the current statements are still relevant. Even though the author was not able to explicitly mention, when a message is encrypted it does not need to be expressed at the routing layer, the SHOULD terms still imply the use of `version` is strongly recommended but is not mandatory. In my opinion this is more of a security consideration that could be added to a new RFC version of the 14/WAKU-MESSAGE if needed.

jm-clius:
@jimstir do I understand correctly that we cannot add to the content at all, including adding a clarifying section where we recommend not using `version`? If that is the case, I think we should consider writing a revised MESSAGE spec with `version` field completely removed, but that is a fairly impactful change as most of our other specs (even stable ones) depend on 14/WAKU-MESSAGE

jimstir:
@jm-clius Correct, the content can not be updated for a `stable` rfc. Even if a new RFC was introduced, it would carry the same title but a new version. So the new RFC would need to become `stable` before replacing the current RFC.

The current RFC could reference the new RFC version.

weboko (approved):
for me looks good!
`version` always seemed excessive

fryorcraken:
Closing for now, as we can do an implementation-only change

PR #176, opened 2025-08-21
docs: add SDS-Repair extension

jm-clius:
Added SDS-R, an (optional) extension of SDS to allow for coordinated repair of missing messages over a limited time window.

It functions by allowing subgroups of participants to rebroadcast dependencies that were reported missing by other participants. As with the rest of SDS, it aims to scale to larger groups first and could likely be simplified for 1:1 or small group chats. To prevent an explosion in repair requests or broadcast storms, it uses backoff timers to prevent multiple participants from performing the same action (either request a repair or rebroadcast in response to a repair request) in a probabilistic manner.

Note that what is still missing is a specified recommended way to use SDS-R in conjunction with retrieving missing dependencies from Store nodes.

### DIFF
--- a/vac/raw/sds.md
+++ b/vac/raw/sds.md
@@ -58,6 +58,9 @@ other participants using the corresponding message ID.
 * **Participant ID:**
 Each participant has a globally unique, immutable ID
 visible to other participants in the communication.
+* **Sender ID:**
+The *Participant ID* of the original sender of a message,
+often coupled with a *Message ID*.
 
 ## Wire protocol
 
@@ -75,6 +78,8 @@ syntax = "proto3";
 message HistoryEntry {
   string message_id = 1; // Unique identifier of the SDS message, as defined in `Message`
   optional bytes retrieval_hint = 2; // Optional information to help remote parties retrieve this SDS message; For example, A Waku deterministic message hash or routing payload hash
+
+  optional string sender_id = 3; //Participant ID of original message sender. Only populated if using optional SDS Repair extension
 }
 
 message Message {
@@ -84,6 +89,10 @@ message Message {
   optional int32 lamport_timestamp = 10;    // Logical timestamp for causal ordering in channel
   repeated HistoryEntry causal_history = 11;  // List of preceding message IDs that this message causally depends on. Generally 2 or 3 message IDs are included.
   optional bytes bloom_filter = 12;         // Bloom filter representing received message IDs in channel
+
+
+  repeated HistoryEntry repair_request = 13; // Capped list of history entries missing from sender's causal history. Only populated if using the optional SDS Repair extension.
+
   optional bytes content = 20;             // Actual content of the message
 }
 ```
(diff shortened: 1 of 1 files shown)


chaitanyaprem (commented):
at a high level the proposal kind of seems like gossipsub-ish (using a cache interval and asking for a group of peers for missing messages etc). 
Wondering if this would require an elaborate method to validate overhead and maybe fine-tune various params to come up with good defaults.

vac/raw/sds.md
```diff
@@ -281,6 +290,176 @@ Upon reception,
...
+
+> **_Note:_** placing `T_resp` values on an exponential backoff curve will likely be more appropriate and is left for a future improvement.
+
+#### Determine response group
+
+Given a message with `sender_id` and `message_id`,
+a participant with `participant_id` is in the response group for that message if
+
+```
+hash(participant_id, message_id) % num_response_groups == hash(sender_id, message_id) % num_response_groups
```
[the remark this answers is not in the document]
jm-clius: I think having groups only formed on participant id could cause hotspots to occur. E.g. if missing messages are heavily skewed towards a specific sender ID we'd expect the same group of participants to always pick up the slack, potentially causing collective overhead in that group that could lead to more losses in the same group, etc. The idea is to spread this responsibility with some pseudorandomisation which is:
- stateless (no need to introduce a "group" field or variable for the message - it can be deterministically computed each time by a participant)
- always includes the original sender id in the response group (as they're most likely to have the message available)

In fact, we can possibly come up with a better algorithm that answers these requirements than hash of participant and message IDs, but this is one simple expression of that.

SionoiS:
SDS is good at identifying missing messages but finding or delivering them is another matter.

TBH if all participants are online why not just sync? Much simpler. If they are not you can't get the message anyway so... :shrug:

jm-clius:
> i am wondering if we can use Waku-sync between 2 participants itself once some missing messages are identified

Yes, this would be what an eventual p2p sync mechanism should allow. Note however that SDS functions only _within_ a logical group conversation over a _broadcast_ transport (like a single Waku pubsub topic). Any p2p syncing (even if logically only between 2 participants) will be broadcast to everyone, which breaks scalability. Participants within an application does not by themselves know how to create a direct connection to another participant. The way to go about this is to either add some negotiation of a direct connection (RTC?) between two participants (using the broadcast transport network for signalling) or simulate something akin to this by having short-lived pubsub topics temporarily used by two participants while syncing.

vac/raw/sds.md
```diff
@@ -281,6 +290,176 @@ Upon reception,
...
+`T_resp` is computed as a pseudorandom backoff from the timestamp when the repair was first requested.
+[Determining `T_resp`](#determine-t_resp) is described below.
+We describe below how a participant can [determine if they're an eligible responder](#determine-response-group) for a specific repair request.
+- Augmented local history log:
+for each message ID kept in the local log for which the participant could be a repair responder,
+the full SDS `Message` must be cached rather than just the message ID,
+in case this participant is called upon to rebroadcast the message.
+We describe below how a participant can [determine if they're an eligible responder](#determine-response-group) for a specific message.
+
+**_Note:_** The required state can likely be significantly reduced in future by simply requiring that a responding participant should _reconstruct_ the original `Message` when rebroadcasting, rather than the simpler, but heavier, requirement of caching the entire received `Message` content in local history.
```
fryorcraken, line 342: Will that be possible? as a SDS message id is a hash, how can one reconstruct?

fryorcraken (commented):
Partial review, will resume later.

vac/raw/sds.md
```diff
@@ -281,6 +290,176 @@ Upon reception,
...
+
+Given a message with `sender_id` and `message_id`,
+a participant with `participant_id` is in the response group for that message if
+
+```
+hash(participant_id, message_id) % num_response_groups == hash(sender_id, message_id) % num_response_groups
+```
+
+where `num_response_groups` is as set out in [SDS-R global state](#sds-r-global-state).
+This ensures that a participant will always be in the response group for its own published messages.
```
fryorcraken, line 447: May be interesting to state that this is a good logic because it allows a participant to determine upon receiving a message, whether they are in the response group (assuming `num_response_groups` doesn't change).

So that it can discard messages (after application processing) the participant is not in the response group, on first reception of said message.

vac/raw/sds.md
```diff
@@ -281,6 +290,176 @@ Upon reception,
...
+
+Furthermore, to avoid a broadcast storm with multiple participants responding to a repair request,
+participants in a single channel MAY be divided into discrete response groups.
+Participants will only respond to a repair request if they are in the response group for that request.
+The global `num_response_groups` variable configures the number of response groups for this communication.
+Its use is described below.
+A reasonable default value for `num_response_groups` is one response group for every `128` participants.
+In other words, if the (roughly) expected number of participants is expressed as `num_participants`, then
+`num_response_groups = num_participants div 128 + 1`.
+In other words, if there are fewer than 128 participants in a communication,
```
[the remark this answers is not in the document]
fryorcraken: I think we need to make a note on whether this value can be `dynamic` and what an implementer should watch out for if they decide to make it dynamic.

vac/raw/sds.md
```diff
@@ -281,6 +290,176 @@ Upon reception,
...
+For each item in the buffer,
+the participant SHOULD broadcast the corresponding `Message` from local history
+if its corresponding response timestamp, `T_resp`, has expired (in other words, `T_resp <= current_time`).
+
+#### SDS-R Periodic Sync Message
+
+If the participant is due to send a periodic sync message,
+it SHOULD send the message according to [SDS-R send message](#sds-r-send-message)
+if there are any eligible items in the outgoing repair request buffer,
+regardless of whether other participants have also recently broadcast a Perdiodic Sync message.
```
fryorcraken, line 461:
```suggestion
regardless of whether other participants have also recently broadcast a Periodic Sync message.
```
(claude)

vac/raw/sds.md
```diff
@@ -281,6 +290,176 @@ Upon reception,
...
+  It MUST compute the `T_resp` for each such repair request according to the steps outlined in [_Determine T_resp_](#determine-t_resp).
+  It MUST determine if it's an eligible responder for a repair request according to the steps outlined in [_Determine response group_](#determine-response-group).
+
+#### Determine T_req
+
+A participant determines the repair request timestamp, `T_req`,
+for a missing `HistoryEntry` as follows:
+
+```
+T_req = current_time + hash(participant_id, message_id) % T_max + T_min
```
fryorcraken, line 402: should we make a recommendation on the `hash` function to use?

(claude)

fryorcraken: (from claude, sounds correct to me)

The formula T_req = current_time + hash(participant_id, message_id) % T_max + T_min has an off-by-one error. Should be % (T_max - T_min) + T_min to
    properly distribute between T_min and T_max


  The current formula is:
  T_req = current_time + hash(participant_id, message_id) % T_max + T_min

  Let's trace through with example values:
  - T_min = 30 seconds
  - T_max = 120 seconds
  - We want delays between 30 and 120 seconds

  With the current formula:
  - hash() % T_max gives values from 0 to 119
  - Adding T_min gives: 30 to 149 seconds
  - Problem: Maximum delay is 149 seconds, not 120 seconds

  The corrected formula should be:
  T_req = current_time + hash(participant_id, message_id) % (T_max - T_min) + T_min

  With the corrected formula:
  - hash() % (T_max - T_min) gives 0 to 89
  - Adding T_min gives: 30 to 119 seconds
  - Correct: Delays are properly bounded between T_min and T_max

  The issue is that the current formula treats T_max as the modulo operand rather than as the upper bound of the desired range. This creates delays that
  can exceed T_max, which violates the stated intention that "T_req to be pseudorandomly and linearly distributed as a backoff of between T_min and
  T_max".

vac/raw/sds.md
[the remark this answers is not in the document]
fryorcraken: (from claude)

- How is num_response_groups synchronized across participants?
- What happens if participants have different values?

Synchronization assumptions: Protocol assumes synchronized T_min, T_max, and num_response_groups without specifying how.

(I think we should comment on that)

vac/raw/sds.md
fryorcraken, line 1: (from claude, a general comment on the buffers in the specs)

Augmented local history:
    - No guidance on retention period for cached messages
    - Memory implications unclear for high-volume channels

Buffer management:
    - No specified limits for repair request/response buffers
    - Could lead to unbounded memory growth

Memory exhaustion attack: Malicious participant could flood repair requests, causing unbounded buffer growth.

(I think it may be good to actually propose some upper bound to all our buffers)

fryorcraken:
It would be good to have a test vector for all the various operations: xor, hash, combining. To ensure all implementations are aligned.

PR #187, opened 2025-09-19
New RFC: CODEX DHT

jimstir:
A new rfc for the Codex DHT component.


kmk142789 (approved):

kmk142789 (approved):

emizzle (approved):
In general, this is good, thank you @jimstir! I noticed several spelling/grammatical errors, and instead of pointing each one out, it's probably easiest to run spellcheck on this document.

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+The `bitsPerHop` MUST indicate the minimum number of bits of a `NodeId` needed to get closer to finding the target per query.
+Practically, it tells a `provider` also how often the node "not in range" branch will split off.
+Setting this value to 1 is the basic, non accelerated version,
+which will never split off the not in range branch and
+which will result in $ \log_2 n $ hops per lookup.
+Setting it higher will increase the amount of splitting on a "not in range" branch,
+thus holding more `providers` with a better keyspace coverage and
+will result in an improvement of $ \log_{2^b} n $ hops per lookup.
+
+- `DistanceCalculator`: value MUST be generated with
```
fbarbu15: should there be more text here? the sentence seems to be cut
the DiscV5 defines the XOR/log2 algorithm explicitly

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+This document explians the Codex DHT (Data Hash Table) component.
+It is used to store Codex's signed-peer-record (SPR) entries,
+as well as content identifiers (CID) for each host.
+
+## Background
+
+Codex is a network of nodes, identified as providers,
+particapting in a decentralized peer-to-peer storage protocol.
+The decentralized storage solution offers data durability guarantees,
+incentive mechanisms and data persistenace guarantees.
```
gmega: Maybe run a spellchecker?

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+   "Provider" : {
+        id: NodeId
+        pubkey: PublicKey
+        address: Array[peerId]
+        record: SignedPeerRecord
+        seen: float // Indicates if there was at least one successful
+        // request-response with this provider, or if the node was verified
+        // through the underlying transport mechanisms. After first contact
+        // it tracks how reliable is the communication with the provider.
+        stats: Stats # traffic measurements and statistics
```
gmega: I've got a few questions here:
1. why do we send an "address" array if SignedPeerRecord already contains a sequence of addresses?
2. Why do we send what would appear to be locally computed statistics, useful to the local peer only, to other peers in the reply?

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+   "IpLimits" : {
+      "tableIpLimit": number,
+      "bucketIpLimit": number
+   }
+
+
+```
+
+The `bitsPerHop` MUST indicate the minimum number of bits of a `NodeId` needed to get closer to finding the target per query.
+Practically, it tells a `provider` also how often the node "not in range" branch will split off.
```
gmega: I have no idea what this means. :-) What's a "not in range" branch?

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+When contructing the `NodeId`,
+the `provider` MUST use the Keccak256 hash function of it's `PublicKey`.
+
+### Signed Peer Record
+
+The `record` MUST be generated by the `provider`,
+which contains their connection information.
+On the Codex network,
+the `record` is identitfied a `SignedPeerRecord`, SPR.
+The strucutue of an SPR is modified from the ENR,
```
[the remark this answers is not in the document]
jimstir: 👍 Removed this statement.

codex/raw/dht.md
jimstir: This was provided through the Codex requirement docs. What do you mean by sequence of addresses included in `SignedPeerRecord`? Also I updated to include the `Stats` object definition.

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+A `provider` SHOULD disregard messages from a node if the `record` is unsigned or become stale.
+The `provider` SHOULD contact other live nodes to disseinate new and updated records.
+The update will increase the `seqNum` then sign the new version of the `record`.
+
+### Distance calculation
+
+#### Routing table
+
+Each `provider` has a local routing table which stores the SPR of other nodes in the netwokr.
+The routing table is accessiable by any node who knows the identity of the `provider`.
```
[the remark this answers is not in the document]
jimstir: Yes these were not to clear, here the updated version:

"Each `provider` has a local routing table which stores the SPR of other `provider`s it has discovered.
New `provider`s SHOULD query live nodes to update their local routing table with SPRs."

codex/raw/dht.md
jimstir: Here's the update statement:
"Practically, it tells a `provider` also how often a node "not in range", based on `NodeId` prefix similarities,
will cause a branch to split off."

codex/raw/dht.md
```diff
@@ -0,0 +1,178 @@
...
+and on the individual buckets.
+In each case, the active node entries,
+but also the entries waiting in the replacement cache are accounted for.
+This way, the replacement cache can't get filled with nodes that then can't be added due to the limits that apply.
+As entries are not verified immediately before or on entry,
+it is possible that a malicious node could fill the routing table or
+a specific bucket with SPRs that have `ip`s it does not control.
+This would effect the node that actually owns the `ip`,
+as they could have a difficult time getting its SPR distrubuted in the DHT.
+However, that `provider` can still search and find nodes to connect to.
```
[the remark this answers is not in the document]
jimstir: Heres an updated:

"As entries are not verified immediately before or on entry,
a malicious node MAY fill the routing table or
a specific bucket with SPRs that have `ip`s not controlled by that adversary.
This would affect the `provider` that actually controls the `ip`,
as they could have a difficult time disseminating its SPR to be stored in the DHT by other `provider`s.
However, that `provider` can still search and find nodes to connect to."

codex/raw/dht.md
jimstir: Yes, sorry I missed this. I thought it would be best to reference the discv5 spec here. here the updated statement:

"- `DistanceCalculator`: value SHOULD be generated with the defined node distance algorithm used in the [discv5 specification](https://github.com/ethereum/devp2p/blob/master/discv5/discv5.md)."

PR #191, opened 2025-09-26
New RFC: CODEX-MANIFEST

jimstir:
RFC for Codex manifest....

### DIFF
--- a/codex/raw/manifest.md
+++ b/codex/raw/manifest.md
@@ -71,8 +71,8 @@ Message Manifest {
   optional uint32 codec = 4;         # Dataset codec
   optional uint32 hcodec  = 5        # Multihash codec
   optional uint32 version = 6;       # Cid version
-  optional string filename = 8;      # original filename
-  optional string mimetype = 9;      # original mimetype
+  optional string filename = 7;      # original filename
+  optional string mimetype = 8;      # original mimetype
 }
 
 ```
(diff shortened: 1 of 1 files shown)


codex/raw/manifest.md
```diff
@@ -0,0 +1,94 @@
...
+instead of hashes of individual pieces,
+we create a Merkle Tree computed over the blocks in the dataset.
+We then include the CID of the root of this Merkle Tree as treeCid attribute in the Codex Manifest file.
+
+In version 2 of the BitTorrent protocol also uses Merkle Trees and
+includes the root of the tree in the info dictionary for each .torrent file.
+
+The Codex manifest, CID in particular,
+is the ability to uniquely identify the content and
+be able to retrieve that content from any single Codex client.
```
emizzle: Why not use the [Rationale section](https://github.com/codex-storage/codex-docs-obsidian/blob/main/10%20Notes/Specs/Codex%20Manifest%20Spec.md#rationale) from the component spec instead? These paragraphs lack depth that is presented in the component spec.

codex/raw/manifest.md
```diff
@@ -0,0 +1,94 @@
...
+  optional uint32 version = 6;       # Cid version
+  optional string filename = 7;      # original filename
+  optional string mimetype = 8;      # original mimetype
+}
+
+```
+
+| attribute | type | description |
+|-----------|------|-------------|
+| `treeCid` | string | A hash based on [CIDv1](https://github.com/multiformats/cid#cidv1) of the root of the [Codex Tree], which is a form of a Merkle Tree corresponding to the dataset described by the manifest. Its multicodec is (codex-root, 0xCD03) |
```
fbarbu15: protobuf shows this as bytes, also the codex specs as well

codex/raw/manifest.md
```diff
@@ -0,0 +1,94 @@
+---
+title: CODEX-MANIFEST
+name: Codes Manifest
```
fbarbu15: typo? Codes

fbarbu15 (approved):

Cofson:
@jimstir 

- Protobuf syntax seems inconsistent in terms of comments especially - using // sometimes for comments and # in other cases. Take a closer look. Should use // for comments. Also shouldn't `mimetype` be string not integer?
- Could reference Codex Merkle tree as it is mentioned but not explained, just for clarity.
- There are some validation rules missing, but not of highest priority right now.
- Some security considerations, but I wouldn't want to go there right now, perhaps it will be fixed with the audit. Not prio right now I think, just to be aware. 
- Idea: could this potentially be integrated with Erasure Coding?

PR #193, opened 2025-09-29
VAC/RAW/ ETH-MLS-OFFCHAIN RFC multi-steward support

seugu:
done: 
- adding multi steward support sections such as: consensus types, big and small type flows
- adding new two de-mls term epoch and backup steward
- adding violation list section

### DIFF
--- a/vac/raw/eth-mls-offchain.md
+++ b/vac/raw/eth-mls-offchain.md
@@ -323,7 +323,7 @@ therefore each member can verify that there is no `MLS proposal` corresponding t
 If there is a steward who releases a commit message without it in the steward list or in the wrong order,
 this is counted as unauthorized steward activity resulting in emergency criteria consensus.
 
- ## Security Considerations
+## Security Considerations
 
 In this section, the security considerations are shown as de-MLS assurance.
 
--- a/vac/raw/eth-mls-offchain.md
+++ b/vac/raw/eth-mls-offchain.md
@@ -293,7 +293,7 @@ which MAY not be suitable for the big groups with more than 1000 members.
 
 ### Multi steward with small consensuses
 
-The two-consensus model offers improved efficiency with a trade-off in decentralization.
+The small-consensus model offers improved efficiency with a trade-off in decentralization.
 In this design, group changes require consensus only among the stewards, rather than all members.
 Regular members participate by periodically selecting the stewards but do not take part in each decision.
 This structure enables faster coordination since consensus is achieved within a smaller group of stewards.
--- a/vac/raw/eth-mls-offchain.md
+++ b/vac/raw/eth-mls-offchain.md
@@ -291,7 +291,7 @@ A big consensus provides better decentralization,
 but it requires a big consensus,
 which MAY not be suitable for the big groups with more than 1000 members.
 
-### Multi steward with two consensuses
+### Multi steward with small consensuses
 
 The two-consensus model offers improved efficiency with a trade-off in decentralization.
 In this design, group changes require consensus only among the stewards, rather than all members.
--- a/vac/raw/eth-mls-offchain.md
+++ b/vac/raw/eth-mls-offchain.md
@@ -97,9 +97,17 @@ for new joiners and new entropy for removed members. In this RFC, the committers
 
 ### de-MLS Objects
 
-`Voting proposal` Similar to MLS proposals, but processed only if approved through a voting process.
+`Voting proposal`: Similar to MLS proposals, but processed only if approved through a voting process.
 They function as application messages in the MLS group,
 allowing the steward to collect them without halting the protocol.
+
+`Epoch steward`: The steward assigned to commit in epoch E according to the steward list.
+Holds the primary responsibility for creating the steward commit in that epoch.
+
+`Backup steward`: The steward next in line after the epoch steward in epoch E.
+Only becomes active if the epoch steward is malicious or fails,
+in which case it completes the commitment phase.
+If unused in epoch E, it automatically becomes the epoch steward in epoch E+1.
 
 ## Flow
 
@@ -211,7 +219,47 @@ Thus, the multi-steward approach primarily defines how steward roles
 rotate across epochs while preserving the underlying structure and logic of the original protocol.
 Two variants of the multi-steward design are introduced to address different system requirements.
 
-### Multi steward with single consensus
+### Consensus Types
+
+Consensus is agnostic with its payload, threfore it can be used for different purposes.
+It is used in three ways here:
+
+1. `Commit proposal:` Each commit message MUST be committed with its YES vote by a specific steward.
+`Commit proposal` is the proposal instance that is specified in Creating Voting Proposal section
+with Proposal.payload MUST show the commit request from `members`.
+2. `Steward election proposal:` Each steward list MUST be verified by this consensus;
+therefore, each member knows and identifies which epoch will be in charge
+and if there is an unauthorised steward commit.
+Proposal.payload MUST represent the identities of stewards in an order.
+3. `Emergency criteria proposal:` If there is a malicious member or steward,
+this event MUST be voted on to finalize it.
+If this returns YES, the next epoch MUST include the removal of the member or steward.
+Proposal.payload MUST consist of the evidence of the dishonesty as described in the Steward violation list.
+
+### Steward list
+
+A steward list is an ordered list that consists of authorized stewards who are eligible
+to create commits when a particular steward commit order should be defined beforehand.
+Therefore, if a malicious steward occurred, the backup steward will be charged with committing.
+
+The steward list is an ordered list; the size is determined when the group is created.
+The index of the slots shows that the Epoch and Steward ID are stored.
+The next in line steward for the epoch E is named as epoch-steward, which has index E.
+And the subsequent steward in the epoch E is named as the backup steward.
+
+If the epoch steward is honest, the back-up steward does not involve the process in epoch,
+and the back-up steward will be the epoch steward within the epoch E+1.
+
+If the epoch steward is malicious, the backup steward is involved in the commitment phase in epoch E
+and the next steward becomes the backup steward in epoch E.
+
+Liveness related to the steward list: After the Steward list is done,
+the members proceed with another set of stewards, which could be the same set,
+then call the consensus type 2.
+This had priority against the application of messaging,
+so till the new set of Stewards are voted as YES, the application is frozen.
+
+### Multi steward with big consensus
 
 In this model, all group modifications, such as adding or removing members,
 must be approved through consensus by all participants,
@@ -222,6 +270,27 @@ However, this benefit comes with reduced operational efficiency.
 The model is therefore best suited for small groups that value
 decentralization and censorship resistance more than performance.
 
+To create a multi-steward with a big consensus,
+the group is initialized with a single Steward as specified as follows:
+
+1. The steward initialized the group with the config file.
+This config file MUST contain `sn` as the steward list size.
+2. The steward adds the members as a centralized way till the number of members reaches the `sn`.
+After the number of members reaches `sn`, members propose lists by voting proposal
+as a consensus among all members, as mentioned in the consensus section 2, according to the checks:
+the size of the voting proposal is equal to the `sn`.
+3. After the voting proposal ends up with a steward list, the freezing also ends,
+and application message is done and group changes committed as specified in single steward section
+with a difference which is members also checks the comittted steward is epoch steward or back-up steward,
+otherwise any one can create the third type of consensus and waiting for the response for emergency call.
+4. If the epoch steward violates the changing process as mentioned in the section Steward violation list,
+one of the members MUST initialize the consensus section 3 to remove the malicious Steward.
+Also, the backup steward fulfills the epoch by committing again correctly.
+
+A big consensus provides better decentralization,
+but it requires a big consensus,
+which MAY not be suitable for the big groups with more than 1000 members.
+
 ### Multi steward with two consensuses
 
 The two-consensus model offers improved efficiency with a trade-off in decentralization.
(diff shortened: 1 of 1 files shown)


vac/raw/eth-mls-offchain.md
```diff
@@ -90,28 +94,45 @@ Here, the add and remove proposals are used.
...
 ### de-MLS Objects
 
-`Voting proposal` Similar to MLS proposals, but processed only if approved through a voting process.
+This section presents the de-MLS objects:
+
+`Voting proposal`: Similar to MLS proposals, but processed only if approved through a voting process.
 They function as application messages in the MLS group,
 allowing the steward to collect them without halting the protocol.
+There are three types of `voting proposal` according to the type of consensus as in shown Consensus Types section,
+these are, `commit proposal`, `steward election proposal` and `emercency criteria proposal`.
```
seemenkina:
```suggestion
these are, `commit proposal`, `steward election proposal` and `emergency criteria proposal`.
```

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +225,156 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+
+This means that if a higher-priority consensus proposal is present in the network,
+lower-priority messages MUST be withheld from transmission until the higher-priority proposals have been finalized.
+
+### Steward list creation
+
+The `steward list` size as `sn` is determined when the group is created.
+The index of the slots shows that the Epoch and `member id`s are stored.
+The next in line steward for the `epoch E` is named as `epoch steward`, which has index E.
+And the subsequent steward in the `epoch E` is named as the `backup steward`.
```
seemenkina: little bit hard to understand, maybe example will help

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +225,156 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+## Steward violation list
+
+A steward’s activity is called a violation if the action is one or more of the following:
+
+1. Broken commit: The steward releases a different commit message from the voted `commit proposal`.
+This activity is identified by the `members` since the [MLS RFC 9420](https://datatracker.ietf.org/doc/rfc9420/) provides the methods
+that members can use to identify the broken commit messages that are possible in a few situations,
+such as commit and proposal incompatibility. Specifically, the broken commit can arise as follows:
+The commit belongs to the earlier epoch. The commit message should equal the latest epoch,
+or the commit needs to be compatible with the previous epoch’s `MLS proposal`.
```
seemenkina:
```suggestion
    1. The commit belongs to the earlier epoch. 
    2. The commit message should equal the latest epoch
    3. The commit needs to be compatible with the previous epoch’s `MLS proposal`.
```

vac/raw/eth-mls-offchain.md
```diff
@@ -90,28 +94,45 @@ Here, the add and remove proposals are used.
...
+`Epoch steward`: The steward assigned to commit in `epoch E` according to the steward list.
+Holds the primary responsibility for creating commit in that epoch.
+
+`Backup steward`: The steward next in line after the `epoch steward` on the `steward list` in `epoch E`.
+Only becomes active if the `epoch steward` is malicious or fails,
+in which case it completes the commitment phase.
+If unused in `epoch E`, it automatically becomes the `epoch steward` in `epoch E+1`.
+
+`Steward list`: It is an ordered list that consists of authorized stewards' `member id`s which belongs
+to stewards who eligible to create commits when a particular steward commit order should be defined beforehand.
```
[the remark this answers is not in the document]
seugu: Indeed, I clarified and gave a mini example [here](https://github.com/vacp2p/rfc-index/pull/193/commits/35e1610966acf2469bc836c40d7f7c2798df5749) and [here](https://github.com/vacp2p/rfc-index/pull/193/commits/de5241cccc657c88c1d4dbf3b43a3be85abf519d)

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +225,156 @@ including welcome messages for the new members. Therefore, the `commit message`
...
-Two variants of the multi-steward design are introduced to address different system requirements.
+Two variants of the multi steward design are introduced to address different system requirements.
+
+### Consensus Types
+
+Consensus is agnostic with its payload; therefore, it can be used for various purposes.
+Note that each message for the consensus of proposals is an `application message` in the MLS object section.
+It is used in three ways as follows:
+
+1. `Commit proposal`: Each commit message MUST be committed with its YES vote by a specific steward.
```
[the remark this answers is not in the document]
seugu: You know, in single steward we had only one steward and updates all commit messages when the proposal is voted YES right. In here, it is the same, only difference is steward must be epoch steward. Otherwise, if we have many stewards that want to commit, I am considering that might be conflicts. 
[Here is a clarification](https://github.com/vacp2p/rfc-index/pull/193/commits/23d31f128a11dd060aedcfcb352da8d99c990568).

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,162 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+
+A `Steward election proposal` is considered valid only if the resulting `steward list`
+is produced through a deterministic process that ensures an unbiased distribution of steward assignments,
+since allowing bias could enable a malicious participant to manipulate the list
+and retain control within a favored group for multiple epochs.
+
+The list MUST be composed of the first `sn` members from the member list,
+sorted according to the ascending value of `SHA256(epoch E || member id)`,
+where `epoch E` is the epoch in which the election proposal is initiated.
+Any proposal with a list that does not adhere to this generation method MUST be rejected by all members.
```
jazzz: I love how this is deterministic as it means every client can derive the same steward list. However I'm curious about the case where the current steward list contains only "Stale" members. (Members who are are valid members of the group, but remain offline for an extended period of time; Either due to unavailability, or a fault causing a desynchronization in state)
- The larger the group the more likely that stale clients exist
- During this time proposals build up, but there is no steward available to generate valid commits.
- Lack of commits results in a partial deadlock, until the era expires and a new steward election proposal occurs and an "online" steward is selected.

Is the intention that applications must prune inactive members to avoid this case?

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,162 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+(which MAY include some or all of the previous members).
+This election is conducted through a type 2 consensus procedure, `steward election proposal`.
+
+A `Steward election proposal` is considered valid only if the resulting `steward list`
+is produced through a deterministic process that ensures an unbiased distribution of steward assignments,
+since allowing bias could enable a malicious participant to manipulate the list
+and retain control within a favored group for multiple epochs.
+
+The list MUST be composed of the first `sn` members from the member list,
+sorted according to the ascending value of `SHA256(epoch E || member id)`,
```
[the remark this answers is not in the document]
jazzz: Not critical, but worth noting. 
Using only static data to derive the list means that malicious actors could synthesis a set of member_ids which provide many epochs of stewardship in the future. I don't think incorporating ratcheted data is worth the complexity at this point; but worth noting in the security model.

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,162 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+which sets and orders stewards responsible for creating commits over a predefined number of epochs.
+This proposal is created periodically, once every number of epochs equal to the steward list size,
+and SHOULD be initialized by any member during the corresponding epoch.
+The Proposal.payload field MUST represent the ordered identities of the proposed stewards.
+Each steward election proposal MUST be verified and finalized through the consensus process
+so that members can identify which steward will be responsible in each epoch
+and detect any unauthorized steward commits.
+3. `Emergency criteria proposal`: If there is a malicious member or steward,
+this event MUST be voted on to finalize it.
+If this returns YES, the next epoch MUST include the removal of the member or steward.
```
jazzz: >the next epoch MUST include the removal of the member or steward.
Is this removal from the group or from the steward list?

If Steward List: Won't that create ambiguity over who the steward is supposed to be?
If Group: Inactivity is listed as reason 3 in the Steward violation list. If stewards are randomly assigned, this means that members could be removed from groups due to network connectivity issues which has usability and security implications

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,162 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+Consensus is agnostic with its payload; therefore, it can be used for various purposes.
+Note that each message for the consensus of proposals is an `application message` in the MLS object section.
+It is used in three ways as follows:
+
+1. `Commit proposal`:  It is the proposal instance that is specified in Creating Voting Proposal section
+with `Proposal.payload` MUST show the commit request from `members`.
+Any member MAY create this proposal in any epoch and `epoch steward` MUST collect and commit YES voted proposals.
+This is the only proposal type common to both single steward and multi steward designs.
+2. `Steward election proposal`: This is the process that finalizes the `steward list`,
+which sets and orders stewards responsible for creating commits over a predefined number of epochs.
```
jazzz: If a steward is removed due to an emergency proposal, what happens to the steward list? Does a new steward election happen?

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,162 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+
+If the `epoch steward` is honest, the `backup steward` does not involve the process in epoch,
+and the `backup steward` will be the `epoch steward` within the `epoch E+1`.
+
+If the `epoch steward` is malicious, the `backup steward` is involved in the commitment phase in `epoch E`
+and the former steward becomes the `backup steward` in `epoch E`.
+
+Liveness criteria:
+
+Once the active `steward list` has completed its assigned epoch,
```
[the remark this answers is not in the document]
seugu: Yes, in fact it is a range starts, like a bijective map with epochs and stewards. In here, I wanted to say "practically ends with last epoch as a single one" but this works also.

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +225,156 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+
+- `Emergency criteria proposal`
+
+- `Steward election proposal`
 
-### Multi steward with single consensus
+- `Commit proposal`
+
+This means that if a higher-priority consensus proposal is present in the network,
+lower-priority messages MUST be withheld from transmission until the higher-priority proposals have been finalized.
```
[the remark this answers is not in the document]
seugu: Thank you for very detail proposals. let's have another round then consensus on the mechanism then I can add them as another section. 

> All should stewards collect incoming proposal

It is a bit hard not only the decryption, also the all stewards needs to be online all the time. Also we might assume some stewards can be offline or network break since our assumption is this works on unreliable network. But, we can make it, few stewards can get keyPackages `epoch `and `backup stewards`. Since, the non-member nodes can see the public key of stewards in the current list in welcome topic, then encrypt then send the keyPackages these two steward.

> If we get emergency proposal, we move group to Waiting State 

I quite like this idea, we can let all know about the waiting state with its reason, since the reason is transparent all the group I think this action is verifiable. 

> Q: who should send emergency proposal in case that steward for example just have a bad connection and can't create commit right now. Do we have any solution in waku for such ping?

Waku has SDS but not sure %100 but Waku has ping since it is build on top of nim libp2p and nim libp2p has default action as mostly for other p2p protocols. I think, we do not need to restrict the sender or emergency proposal everyone can init it, we can maybe say that the application can restrict instead of emergency proposal DoS. 

> if we previously have some proposals from this steward , do we reject it or no?

I think, previous proposals can stay since at that time this steward was honest, but it is not now.

> after everyone get it we change state from emergency to waitnig and start usual epoch, so that means we handle all previously collected proposals. And here we perhaps can skip time gap for emergency if we take to much

This is very cool, it means we just transfer the proposals to next epoch so we do not need to reject them.

> steward not create commit when new epoch starts because he is offline - we need additional timeout in epoch to understand that steward not response and make next steward as main. Maybe even not to kick this user from the group

It is easy to init a emergency proposal when there is concrete violation but, for inactivity, there is a [comment ](https://github.com/vacp2p/rfc-index/pull/193#discussion_r2449137250) about the time-epoch issue. I think we have no exact time interval for epoch right such as in seconds. The epoch is changed when the steward release a commit message. So, we need to have some time interval to define the inactivity situtation somehow, otherwise how we can understand that the steward is silent (malicious or network issue does not matter).

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,173 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+(which MAY include some or all of the previous members).
+This election is conducted through a type 2 consensus procedure, `steward election proposal`.
+
+A `Steward election proposal` is considered valid only if the resulting `steward list`
+is produced through a deterministic process that ensures an unbiased distribution of steward assignments,
+since allowing bias could enable a malicious participant to manipulate the list
+and retain control within a favored group for multiple epochs.
+
+The list MUST be composed of the first `sn_min` members including previous stewards,
+sorted according to the ascending value of `SHA256(epoch E || member id || group id)`,
```
jazzz: I think this section could use some added clarity, to make it super clear to readers.

Specifically that it is made by computing a hash **for each member of the group**, and sorting the results into the steward list.

vac/raw/eth-mls-offchain.md
```diff
@@ -204,32 +231,173 @@ including welcome messages for the new members. Therefore, the `commit message`
...
+where `epoch E` is the epoch in which the election proposal is initiated,
+and `group id` for shuffling the list across the different groups.
+Any proposal with a list that does not adhere to this generation method MUST be rejected by all members.
+
+We assume that there are no recurring entries in `SHA256(epoch E || member id)`, since the SHA256 outputs are unique
```
jazzz: update to match above

```suggestion
We assume that there are no recurring entries in `SHA256(epoch E || member id || group id)`, since the SHA256 outputs are unique
```

PR #195, opened 2025-10-01
Fix SDS lamport timestamp overflow and keep it near current time

fryorcraken:
The SDS timestamp is initialized to the current time in milliseconds, which is a 13 digits value (e.g. 1,759,223,090,052).

The maximum value for int32 is 2,147,483,647 (10 digits), which is clearly less than the timestamp. Maximum value for uint32 is 4,294,967,295 (10 digits), which does not help either.

uint64 is BigInt in JavaScript, so best to be avoided unless strictly necessary as it creates complexity. max uint64 is 18,446,744,073,709,551,615 (20 digits).

Using seconds instead of milliseconds would enable usage of uint32 valid until the year 2106.

The lamport timestamp is only initialized to current time for a new channel. The only scenario is when a user comes in a channel, and thinks it's new (did not get previous messages), and then starts sending messages. Meaning that there may be an initial timestamp conflict until the logs are consolidated, which is already handled by the protocol.

JS implementation: https://github.com/waku-org/js-waku/pull/2664

### DIFF
--- a/vac/raw/sds.md
+++ b/vac/raw/sds.md
@@ -81,7 +81,7 @@ message Message {
   string sender_id = 1;           // Participant ID of the message sender
   string message_id = 2;          // Unique identifier of the message
   string channel_id = 3;          // Identifier of the channel to which the message belongs
-  optional int32 lamport_timestamp = 10;    // Logical timestamp for causal ordering in channel
+  optional uint32 lamport_timestamp = 10;    // Logical timestamp for causal ordering in channel
   repeated HistoryEntry causal_history = 11;  // List of preceding message IDs that this message causally depends on. Generally 2 or 3 message IDs are included.
   optional bytes bloom_filter = 12;         // Bloom filter representing received message IDs in channel
   optional bytes content = 20;             // Actual content of the message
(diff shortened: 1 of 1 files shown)


jm-clius (commented):
Copying from Discord:

> I was even worried that milliseconds might not be a fine enough resolution for ordering when the message rates are high. To explain:
> - using current timestamp to initialise only makes sense if all the lamport timestamps in the channel remain more or less around current time
> - each event increases the lamport timestamp by _at least_ `1`
> - lamport timestamps MUST increase monotonically for a specific user/clock. Since it's only used for ordering it's OK to jump forward with any integer > 0, but never for a specific clock to run backward
> - with second resolution we'll very quickly increase clocks so far into the future compared to real time that someone initialising with current time will have their messages sorted into the distant past (and we could as well have initialised everyone to `0`). With ms resolution even with a 1000 events per second a new joiner's messages wouldn't be more than 1 s (in ordering) in the past.

> Something to check in the implementation(s):
> - the scheme above only works if the lamport timestamps are constantly "corrected" (forwards only) to current time. The way I'd recommend is what Status did in the past, setting the lamport timestamp for _each_ message as:
> `lamport_timestamp = max(current_timestamp, previous_received_lamport_timestamp + 1)`
> I imagine it could also be possible to correct timestamps with a timer event (say once per second), but I think the option above is more elegant.

I've since opened spec change to indicate this: https://github.com/vacp2p/rfc-index/pull/196

fryorcraken:
Superseded by #196

PR #196, opened 2025-10-01
docs: update lamport timestamps to uint64, pegged to current time

jm-clius:
Lamport timestamps should remain close to current time (in milliseconds) for new joiners to be able to have their messages ordered reasonably close to other messages in the channel.

This means that:
- the `timestamp` field should be large enough to accommodate millisecond resolution timestamps, i.e. `uint64` (see https://github.com/vacp2p/rfc-index/pull/195 for reasoning)
- the lamport timestamp should be updated before sending _each_ message to `max(timeNowInMs, current_lamport_timestamp + 1)`.

The current spec only indicated that Lamport timestamps should be _initialised_ to current time, which means that the logical timestamp would soon drift from current time.

### DIFF
--- a/vac/raw/sds.md
+++ b/vac/raw/sds.md
@@ -81,7 +81,7 @@ message Message {
   string sender_id = 1;           // Participant ID of the message sender
   string message_id = 2;          // Unique identifier of the message
   string channel_id = 3;          // Identifier of the channel to which the message belongs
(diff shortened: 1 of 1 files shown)


fryorcraken (approved):

jm-clius:
@shash256 please check implementation in nim-sds

PR #254, opened 2026-01-12
docs: spec for rln membership allocation protocol

adklempner:


vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
+---
+title: RLN-MEMBERSHIP-SERVICE
+name: RLN Membership Allocation Service
```
jm-clius: A bit of a nitpick, but to follow previous conventions (e.g. https://github.com/logos-co/logos-lips/blob/master/docs/messaging/standards/core/12/filter.md), I'd avoid focusing too much on the _service_ aspect here. Although each interaction on the wire has a "service-side" and a "client-side", a specific node can be both service and client, so we don't want to suggest an overly-binary view of request-response protocols. I'd suggest titling this just as `RLN-MEMBERSHIP-ALLOCATION` (or, perhaps, "RLN Membership Allocation Protocol", although in the context of LIPS, the "Protocol" qualifier is a bit redundant).

```suggestion
name: RLN Membership Allocation
```

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+category: Standards Track
+tags: rln
+editor: Arseniy Klempner <arseniyk@status.im>
+contributors:
+---
+
+## Abstract
+
+This specification defines the RLN Membership Allocation
+Service, which registers RLN identity commitments on behalf of clients.
```
jm-clius:
```suggestion
Protocol, which registers RLN identity commitments on behalf of clients.
```
Here and elsewhere - reasoning in my comment above. (This is definitely a nit, so feel free to disagree).

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+The service enables applications to obtain RLN memberships
+without directly interacting with the on-chain membership contract.
```
jm-clius:
```suggestion
The protocol enables membership providers, such as an RLN-controlled application,
to allocate RLN memberships to eligible clients
without requiring those clients to interact directly with the on-chain membership contract.
```
(the applications are most likely to be the providers/sponsors of the memberships, not the recipients).

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+It aims to do so while minimizing linkability between the client's
```
jm-clius: Super nit (here and elsewhere) - we're apparently following British spelling standards, although it's not at all consistently applied and our first specs were in American and not updated yet. Up to you if you want to adapt :)
```suggestion
It aims to do so while minimising linkability between the client's
```

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+
+1. This typically requires clients to have a wallet with sufficient funds to
+pay for gas fees, and access to a node for the blockchain network.
+2. Access to a node is typically done using an RPC provider, which risks
+creating a correlation between the client's IP address and their RLN identity.
+
+The RLN Membership Allocation Service addresses the first concern by acting as an intermediary that registers identity commitments on behalf of clients. Each service provider
+can implement their own authentication mechanism to determine client eligibility for membership allocation.
+
+The second concern can be addressed by using a privacy-preserving identity mechanism like [RLN Stealth Commitments](./rln-stealth-commitments.md), which allows the service to register identity commitments on behalf of clients without gaining knowledge of the client's network identity.
```
jm-clius: Perhaps mention that this is a future concern and out of scope of the current spec?

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+This specification defines the RLN Membership Allocation
+Service, which registers RLN identity commitments on behalf of clients.
+The service enables applications to obtain RLN memberships
+without directly interacting with the on-chain membership contract.
+It aims to do so while minimizing linkability between the client's
+network identity and their RLN identity.
+The specification includes a pluggable authentication mechanism
+to determine client eligibility for membership allocation.
+
+## Motivation
```
jm-clius, line 25: I really liked this motivation section. Thanks!

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Service Provider Eligibility
+
+A service provider SHOULD deposit sufficient funds in a smart contract to cover the cost of registering at least `min_membership_buffer` RLN memberships.
+The service providers smart contract MUST allow depositors to use the deposited funds to register memberships in the RLN membership contract.
+The service providers smart contract MUST allow depositors to withdraw their deposited funds.
+The service providers smart contract MUST track how many memberships have been registered by a depositor.
```
jm-clius: I'm not so sure I get this section. Why do we need a Membership Provider smart contract? Why not simply require that the membership provider has access to transferable funds (probably just its own wallet/account). It should then do local accounting to determine how many memberships it has already submitted and how many it has left. In other words, I don't think this is a concern of the protocol - the only requirement/assumption is that the Membership Provider must be able to submit membership transactions on-chain.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+pay for gas fees, and access to a node for the blockchain network.
+2. Access to a node is typically done using an RPC provider, which risks
+creating a correlation between the client's IP address and their RLN identity.
+
+The RLN Membership Allocation Service addresses the first concern by acting as an intermediary that registers identity commitments on behalf of clients. Each service provider
+can implement their own authentication mechanism to determine client eligibility for membership allocation.
+
+The second concern can be addressed by using a privacy-preserving identity mechanism like [RLN Stealth Commitments](./rln-stealth-commitments.md), which allows the service to register identity commitments on behalf of clients without gaining knowledge of the client's network identity.
+
+## Service Provider Eligibility
```
jm-clius:
```suggestion
## Membership Provider Eligibility
```
Here and elsewhere I'd suggest using a specific role description - "Membership Provider" - rather than the generic "Service Provider". However, I'm not quite convinced that this section is really necessary at all (see my comment below).

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+## Wire Format Specification
+
+### Discovery
+
+A service provider SHOULD advertise that it offers membership allocation services by participating in the [Logos Capability Discovery protocol](./logos-capability-discovery.md).
+A service provider MUST include a list of supported authentication methods in the metadata field of the `Advertisement` message.
+A service provider MUST use the protocol ID `/rln/membership/<version>` when generating the `service_id_hash`.
+A service provider MAY also advertise support for a specific authentication method by using the protocol ID `/rln/membership/<version>/<authentication_method>`.
+
+A client SHOULD send a `GET_ADS` request for the `/rln/membership/<version>` service to registrars to discover service providers that offer membership allocation services.
```
jm-clius: We can either leave Discovery completely out or just add it as an appendix or implementation suggestion. The assumption here is that we have connectivity/discovery set up elsewhere. What _is_ necessary though, is a `libp2p protocol ID` (aka libp2p codec) to indicate what type of libp2p connection should be negotiated and can trigger the correct protocol handlers. In other specs we just add this top-level: https://github.com/logos-co/logos-lips/blob/master/docs/messaging/standards/core/19/lightpush.md?plain=1#L31

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Allocation Request and Response
```
jm-clius: I'd suggest splitting this into two separate sections with the name of the wire message as title for each:
```
### Membership Allocation Request

<insert protobuf and everything you said about the request>

### Membership Allocation Response

<insert protobuf and everything you said about the response>
```

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+    bytes authentication_payload = 2;
+    
+    // The identity commitment to register
+    bytes identity_commitment = 3;
+
+    // Rate limit for the membership
+    uint64 optional rate_limit = 4;
+}
+```
+
```
jm-clius, line 83:
```suggestion

### Membership Allocation Response

```
See my comment above - I think it's more easy to follow in two separate sections.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+
+A client MUST send a message to a service provider to request the allocation of a membership.
```
jm-clius: I like especially my first sentence in the protocol "recipe" to be as exact as possible. Many people will start reading your spec only from here, so they might ask "which clients? what message? to which providers?" We don't have to answer every question in detail, but make sure that they understand there are protocol conditions to be met at each of these points.

For example:
```suggestion
A client who wants to obtain an RLN membership using this protocol,
MUST send a `MembershipAllocationRequest` to a relevant membership provider
to request the allocation of a membership.
```

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+A client MUST specify the authentication method to use in the `authentication_type` field.
+A client MUST include a payload in the `authentication_payload` field, further defined by the authentication method.
```
jm-clius: Wondering if we really need both the generic `authentication_type` and `authentication_payload` field - I'm thinking that if an application requires a specific authentication, this type will already be implicit in how the payload gets encoded. Some providers may indeed want to differentiate between different types of authentication, but even then the type can be encoded into the payload structure.

jm-clius: We should mention here that the authentication is pluggable and up to the specific provider, perhaps adding a clarifying example or two (e.g. some providers may be set up by an application that requires auth tokens from their users to obtain membership, others may be as simple as an app signature to provide short-lived demo memberships, etc.)

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+    // Generic payload, further defined by the authentication mechanism
+    bytes authentication_payload = 2;
+    
+    // The identity commitment to register
+    bytes identity_commitment = 3;
+
+    // Rate limit for the membership
+    uint64 optional rate_limit = 4;
+}
+```
```
jm-clius, line 82: I'd suggest moving the protobuf to before the discussion for this paragraph - I think it's a little bit easier to see the code and then read the instructions on how to use and populate the message.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+    
+    // The identity commitment to register
+    bytes identity_commitment = 3;
+
+    // Rate limit for the membership
+    uint64 optional rate_limit = 4;
+}
+```
+
+A service provider MUST respond to the client's request with a `MembershipAllocationResponse` message.
```
jm-clius:
```suggestion
A membership provider SHOULD respond to a `MembershipAllocationRequest` with a `MembershipAllocationResponse` message.
```
I think there may be a good reason for some providers to just ignore it, hence the SHOULD.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+A service provider SHOULD perform basic validation on the identity commitment, such as total bit length.
+A service provider SHOULD check if the identity commitment has already been registered in the RLN membership contract.
+A service provider SHOULD check if the hash of the request has already been processed in the service provider smart contract.
```
jm-clius: As mentioned elsewhere, I don't think we need a smart contract here - this is simply accounting required of the provider. It _could_ choose to do this in a smart contract, but that seems orthogonal to protocol concerns to me.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+A service provider MUST deserialize the authentication payload based on the authentication method specified in the `authentication_type` field.
+A service provider MUST authenticate the client's request using the authentication payload.
+Upon successful authentication, a service provider SHOULD send a transaction to the RLN membership contract with the `identity_commitment` as an argument.
+A service provider MUST respond with a `MembershipAllocationResponse` message indicating whether the authentication for the request was accepted.
+A service provider MUST respond with a `MembershipAllocationSuccess` message if the transaction to register was successful.
+A service provider MUST respond with a `MembershipAllocationFailure` message if the transaction to register was unsuccessful.
```
jm-clius: Hah. I see. This is a multi-step response with different message types, which will be complex to implement as a libp2p request-response protocol. Usually you can have only one protobuf message for the request and another for the response (otherwise the handler doesn't know how to deserialise what it received). For simplicity, I would advise that the service provider sends only a single `MembershipAllocationResponse` and that this then contains information about the `auth_success` as well as either a nested `MembershipAllocationSuccess` or `MembershipAllocationFailure` message. Of course the response may take a while to wait for the on-chain transaction, but I think it's worth doing for the sake of simplicity (at least for now).

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+### Allocation Request and Response
+
+A client MUST send a message to a service provider to request the allocation of a membership.
+A client MUST specify the authentication method to use in the `authentication_type` field.
+A client MUST include a payload in the `authentication_payload` field, further defined by the authentication method.
+A client MUST include the [identity commitment](../32/rln-v1.md#user-identity) to register in the `identity_commitment` field.
+
+```protobuf
+message MembershipAllocationRequest {
+    // Indicates which authentication mechanism is used
```
jm-clius, line 70: For request response protocols, you generally need a unique `request_id` to correlate specific requests with specific responses (see e.g. pattern [here](https://github.com/logos-co/logos-lips/blob/master/docs/messaging/standards/core/12/filter.md#protobuf)).

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+
+```protobuf
+message MembershipAllocationResponse {
+    // Result of the authentication attempt
+    bool auth_success = 1;
+    uint64 request_id = 2;
```
jm-clius: I can't exactly remember why, but our request IDs are usually `string`. May be a good idea to stick to the convention, although perhaps it wasn't a good convention to begin with. 🤷 I would suggest making the `request_id` tag `1` though, to follow other similar protocols.

jm-clius (changes requested):
Thanks! I added some comments/requests below, many of them quite nitpicky. LMK if you have any questions.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+    bytes authentication_type = 1;
+
+    // Generic payload, further defined by the authentication mechanism
+    bytes authentication_payload = 2;
+    
+    // The identity commitment to register
+    bytes identity_commitment = 3;
+
+    // Rate limit for the membership
+    uint64 optional rate_limit = 4;
```
Cofson: `optional` goes before the type in proto3, so `uint64 optional rate_limit = 4` won't compile. Should be `optional uint64 rate_limit = 4`. Same for `string optional error = 3` in the response message below.

vac/raw/rln-membership-service.md
Cofson: Is `MembershipAllocationResponse` sent first, followed by `Success` or `Failure` as a separate message? Or is the idea to use a `oneof` inside the response? Currently reads like both.

vac/raw/rln-membership-service.md
```diff
@@ -0,0 +1,149 @@
...
+message BasicAuthenticationPayload {
+    bytes message = 1;
+    bytes signature = 2;
+}
+```
+
+Depending on the claims being verified, the authentication payload may require additional fields:
+
+```protobuf
+message BasicAuthenticationPayload {
```
Cofson: `BasicAuthenticationPayload` is defined twice. In one `.proto` file the second definition would shadow the first. Probably want a separate name for the extended version (e.g. `OnchainAuthenticationPayload`).

vac/raw/rln-membership-service.md
adklempner: Here I was thinking ahead a little bit and was considering how the protocol would work with many service providers to choose from; how clients can have some guarantee that the provider is actually capable of fulfilling the request, that a provider has a history of registering memberships. And vice-versa, some indication for providers that they aren't fulfilling a duplicate request. But it's not so simple and I agree that it's out-of-scope

vac/raw/rln-membership-service.md
adklempner: Originally it would first send the `Response`, kind of like an acknowledgement that the request was accepted, followed by a `Success` or `Failure`.

But based on feedback it will now be a single response that uses a `oneof`

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,206 @@
...
+The second concern is out of scope for this specification and is considered
+future work. It can be addressed by using a privacy-preserving identity
+mechanism like [RLN Stealth Commitments](./rln-stealth-commitments.md),
+which would allow the provider to register identity commitments on behalf
+of clients without gaining knowledge of the client's network identity.
+
+## Membership Provider Eligibility
+
+A membership provider SHOULD have access to sufficient funds to cover
+the cost of registering at least `min_membership_buffer` RLN memberships.
```
jm-clius: We need to define this variable (even in line) or perhaps just state it in plain terms:
```A membership provider SHOULD have sufficient funds to register memberships on behalf of clients at the expected rate for the desired application.```

File: docs/ift-ts/raw/rln-membership-service.md [hunk 23]
```diff
@@ -0,0 +1,206 @@
...
+A client MUST include a unique `request_id` to correlate the request with
+the response.
+A client MUST specify the authentication method to use in the
+`authentication_type` field.
+A client MUST include a payload in the `authentication_payload` field,
+further defined by the authentication method.
+A client MUST include the
+[identity commitment](../32/rln-v1.md#user-identity) to register in the
+`identity_commitment` field.
+
```
jm-clius, line 100: For completeness, we should define what goes (optionally) into `rate_limit`.

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,206 @@
...
+Authentication is pluggable: membership providers MAY support any
+authentication mechanism, including token-based authentication,
+cryptographic signatures over eligibility claims, or demo/testnet
+memberships with minimal authentication.
+
```
jm-clius, line 105: I'd add:

```suggestion
A client MAY retry a MembershipAllocationRequest after a suitable timeout if no response is received.
```

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,206 @@
...
+Upon successful authentication, a membership provider SHOULD send a
+transaction to the RLN membership contract with the `identity_commitment`
+as an argument.
+A membership provider's response MUST include the `auth_success` field
+indicating whether the authentication for the request was accepted.
+If the transaction to register was successful, the response MUST include
+a `MembershipAllocationSuccess` message in the `result` field.
+If the transaction to register was unsuccessful, the response MUST include
+a `MembershipAllocationFailure` message in the `result` field.
+
```
jm-clius, line 171:
```suggestion
A membership provider SHOULD include a descriptive error message in `MembershipAllocationFailure`.
```

docs/ift-ts/raw/rln-membership-service.md
```diff
@@ -0,0 +1,206 @@
...
+```protobuf
+message OnchainAuthenticationPayload {
+    bytes message = 1;
+    bytes signature = 2;
+    bytes transaction_hash = 3;
+    uint64 chain_id = 4;
+}
+```
+
```
jm-clius, line 204: This spec is now shaping up so well, we might as well complete it with the standard `Copyright` and `References` section. See e.g. https://lip.logos.co/messaging/standards/core/12/filter.html#copyright

jm-clius (approved):
Last few comments below, but can be merged IMO when done. Thanks!

docs/ift-ts/raw/rln-membership-service.md
adklempner: This was another artifact of the smart contract approach; will go with plain terms

Cofson (approved):

PR #293, opened 2026-02-17
improved Merkle tree spec

bkomuves:
more clarity and context; and an Appendix with concrete hash function instances.


Cofson (approved):
Great work!

If I had to just point one thing, I would suggest to change from using first-person language
throughout ("we propose", "our construction", "let's assume") to passive voice or third person for consistency with the specification writing guidelines.

But this is of lowest priority - I can do it too. As it is right now, this spec, when consensus is reached, can go to stable.

bkomuves:
ok, well, what is the currently PC way to phrase that, "hey, we are doing this seemingly random-looking thingy; but it's most certainly not because that's a goal we want to achieve; you know, it's just randomly appeared from a third person point of view!" ?

mghazwi (commented):
I think this looks great. Much cleaner notation and removes the ambiguity we had before. Thanks. 

I added few comments to fix typos, but no big issues. I think we can merge this.

PR #297, opened 2026-03-18
Restructure spam protection: Split into Exit Abuse Prevention and DoS Protection sections

AkshayaMani:
This PR restructures the spam protection content in `mix.md` for correctness, scope accuracy, and modularity. It reframes exit abuse prevention as a distinct concern, decouples DoS protection into a standalone spec, and integrates both into the relevant packet construction and handler sections. It also includes some formatting and typo fixes and cleanup.

#### Section 6 Pluggable Components

*   **§6.3 Spam Protection → Exit Abuse Prevention**
    *   Full rewrite: scope narrowed to exit nodes only
    *   Renamed to reflect actual mechanism (exit abuse prevention, not general spam protection)
*   **§6.6 DoS Protection** _(new)_
    *   New section pointing to standalone spec (#302 )
*   Renamed all references from spam protection to exit abuse prevention throughout

#### Section 7 Core Mix Protocol Responsibilities

*   **§7.3.2 Exit Processing**
    *   Added exit abuse prevention validation step to exit processing logic

#### Section 8 Sphinx Packet Format

*   **§8.5 Packet Construction**
    *   Updated to embed/parse exit abuse prevention proof in the Sphinx packet

#### Section 9

*   **Remove §9 "Spam Protection Architecture"** in its entirety — content extracted into new spec (separate PR)

### Fixes and Cleanup

*   **§8.6.2 Node Role Determination**
    *   Simplified exit detection condition
    *   Fixed β parsing: removed incorrect dependency on `L`
*   Fix broken section cross-references
*   Formatting fixes and typos

### DIFF
--- a/docs/ift-ts/raw/mix.md
+++ b/docs/ift-ts/raw/mix.md
@@ -359,7 +359,8 @@ Common strategies include Proof of Work (PoW), Verifiable Delay Functions (VDFs)
 The sender is responsible for appending the appropriate exit abuse prevention data (_e.g.,_ nonce, timestamp) to the message payload.
 The format and verification logic depend on the selected method.
 
-Note: The exit abuse prevention mechanisms described above are intended to protect the exit from flooding attacks routed through them.
+Note:
+The exit abuse prevention mechanisms described above are intended to protect the exit from flooding attacks routed through them.
 They do not provide protection against denial-of-service (DoS) or resource exhaustion attacks targeting the mixnet itself (_e.g.,_ flooding mix nodes with traffic, inducing processing overhead, or targeting bandwidth).
 
 Protections against attacks targeting the mixnet itself are addressed in [Section 6.6](#66-dos-protection).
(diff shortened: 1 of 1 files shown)


chaitanyaprem (commented):
Looks Good To me apart from few comments.

Cofson (approved):

AkshayaMani:
> Looking at current intended use-cases of mix i.e logos-messaging and logos-storage and Logos altogether, any node that provides any sort of services will be running mix e.g lightpush in waku, file provider in storage etc. So, maybe it would be better to align the spec with default behavior of `exit==destination`. If this is the case, do we still need to mention exit abuse prevention? IMHO, that can be left to service/application using mix e.g lightpush already has its own spam protection mechanisms. If storage protocols don't need this, maybe we can defer working on this.

These seem like orthogonal concerns. The default being exit is the destination (or not) is a separate decision and shouldn't determine whether exit abuse prevention belongs in the mix spec.
Moreover, the mixnet's unlinkability is what introduces the exit abuse gap. So, it seems natural for the mix spec to at least define a pluggable interface for solving it. Also, having it doesn't exclude application-layer solutions.

jm-clius (approved):
Thanks for this! In general it's a good idea to explicitly re-request review from people that have previously added comments that you've either addressed or answered. :) Let's get this merged, unless there are any major objections re content (minor things can be addressed subsequently). Will be good to get an approval from @chaitanyaprem as well.

chaitanyaprem (approved):
Let's merge it for now. 
As discussed we can get consensus on exit and destination and then make further changes.

PR #299, opened 2026-03-19
Add on-chain part of payment streams spec

s-tikhomirov:
This PR adds a detailed description of the on-chain part of the payment streams protocol, in particular, how the protocol fits into the LEZ model. The spec is being developed alongside the corresponding [PoC implementation](https://github.com/logos-co/logos-lez-payment-streams).


docs/anoncomms/raw/payment-streams.md
```diff
@@ -68,540 +71,889 @@ One such protocol is Store,
...
+The `StreamProposal` MUST satisfy the provider's policy at verification time `t`.
+`owner_signature` MUST verify under `owner_public_key` over the canonical proposal payload.
+The user MUST NOT send another `StreamProposal` for the same vault-provider pair
+while a proposal is pending.
+
+Before serving the request,
+the provider MUST confirm that the proposal satisfies the policy,
+the `VaultProof` requirements,
+and `owner_signature` validity.
+On success it MUST return `EligibilityStatus.status_code` `OK` with `response_data`.
```
jm-clius, line 640: Wtihin a `ServiceResponse`, right? I think it's useful to make this explicit in the flow: essentially the first `ServiceResponse` is free, provided that the proposal and vault proof seems able to back a future stream. Perhaps we can recommend that for this reason the first Response can be limited, depending on what that means for the specific service (e.g. minimal number of entries in Store response).

docs/anoncomms/raw/payment-streams.md
```diff
@@ -68,540 +71,889 @@ One such protocol is Store,
...
+For verification it MUST read that stream's on-chain account state
+for the session `vault_id` under the chain integration,
+and then fold that state to the current chain time.
+
+Each `StreamProof` MUST name the on-chain stream that matches the accepted
+`StreamParams` for the service session,
+and MUST carry a valid signature over `request_data` with the session key.
+The referenced stream MUST be `ACTIVE`.
+On the first valid stream-proof request,
+the provider MUST record `stream_id` in session state if not already set.
```
jm-clius, line 681: It's not entirely clear to me why we need to maintain session state off-chain if every `StreamProof` is accompanied by an on-chain state verification in any case? I can see that we want to prevent the same vault from being used for multiple vault-backed initial requests, but once the stream is created do we need to track anything for the stream session off-chain? Asking, because so far our req-resp services were designed to be stateless, i.e., every request had to contain enough information to be serviced without reference to previous requests.

docs/anoncomms/raw/payment-streams.md
```diff
@@ -68,540 +71,889 @@ One such protocol is Store,
...
+
+When service ends, the user MAY pause or close the on-chain stream
+and MAY stop sending stream proofs.
+
+For `TEMPORARY` termination, the user MAY pause the stream until `resume_after`.
+The provider MAY resume the same service session after `resume_after`.
+
+For `PERMANENT` termination, the user SHOULD close the on-chain stream promptly.
+Further service requires a newly accepted `StreamProposal`.
+
```
jm-clius, line 709: How do claims happen in this flow? It seems from the discussion above, that a claim can only happen on a closed stream? Or could it also happen mid-session once enough funds are accrued to make it worthwhile?

docs/anoncomms/raw/payment-streams.md
s-tikhomirov: Yes, we currently recommend to limit `response_data` to a given number of bytes prescribed in the policy.

docs/anoncomms/raw/payment-streams.md
s-tikhomirov: Claim can be done from any stream state. The provider decides when to claim, and claiming doesn't change the stream state. In case of service termination, the provider can either wait until the user closes the stream and then claim, or claim immeidately (leaving the stream open and then claiming again after the user closes the stream).

docs/anoncomms/raw/payment-streams.md
s-tikhomirov: Great question. There is a trade-off between lightweight request + stateful provider vs heavier requests and stateless provider. The current protocol prefers the former. Shall we address this question in a future version of the protocol? It may indeed make sense to switch to stateless provider (or "nearly" stateless - see below), but that would require updating the flow carefully.

Pasting an AI explanation for more context.

---

## Why off-chain session state is required

The short answer: on-chain verification proves that *some* stream exists and is `ACTIVE`, but it does not prove that *this* request is backed by *the proposal you accepted*, under *the session key the vault owner authorized*, within *the deadline and policy you pinned*. The `StreamProof` is deliberately tiny (just `stream_id` + `signature`), so the context that binds it to an accepted proposal has to live somewhere — and the spec puts it in provider-side session state.

Let me walk through what is actually stored and why each piece cannot be recovered from the chain or from the `StreamProof` alone.

### What the provider records

The session record is defined at acceptance time:

```647:652:rfc-index/docs/anoncomms/raw/payment-streams.md
A service session is the provider's off-chain state for one accepted proposal.
On acceptance the provider MUST record accepted `StreamParams`,
the policy pinned at acceptance,
`StreamProposal.public_key` as the session key,
and `vault_id`, `provider_id`, and `owner_public_key` from the vault proof.
Later stream-proof verification MUST use that pinned policy and session record.
```

Plus `stream_id` is pinned on the first valid stream-proof:

```680:681:rfc-index/docs/anoncomms/raw/payment-streams.md
On the first valid stream-proof request,
the provider MUST record `stream_id` in session state if not already set.
```

So the session record is: `{accepted StreamParams, pinned policy, session public key, vault_id, provider_id, owner_public_key, stream_id}`.

### Why each field is not derivable from the chain or the StreamProof

The `StreamProof` wire shape is minimal by design:

```487:492:rfc-index/docs/anoncomms/raw/payment-streams.md
message StreamProof {
  bytes stream_id = 1;
  bytes signature = 2;
}
```

`signature` is over `request_data` using the committed session key (line 496). The `StreamProof` carries no public key, no params, no policy, no vault id. So:

1. Session key. `signature` can only be checked against `StreamProposal.public_key`, which is *not* on the on-chain `StreamConfig` (only `StreamConfig.provider` is on chain — line 781). The binding "the vault owner authorized this session key" is established by `VaultProof.owner_signature` covering `session_public_key` (lines 533-538). If the `StreamProof` carried the public key itself, the signature would be self-attesting: any key could sign and claim eligibility. The owner-authorization binding lives in the proposal, so the provider must remember which key was authorized.

2. Accepted `StreamParams` and pinned policy. The on-chain stream stores `rate`, `allocation`, `provider`, `accrued`, etc. It does *not* store `service_id`, `create_stream_deadline`, or which policy version was satisfied at acceptance. Policy advertisements update over time, but "updated policy advertisements apply only to new proposals" (line 626). Matching is defined as exact equality against the *accepted* params (lines 560-562, 664-665, 676-677), so the provider needs the accepted params, not just the current on-chain fields or the current policy.

3. `vault_id` / `owner_public_key`. On LEZ, `StreamConfig` is a PDA derived from the `VaultConfig` address plus the stream id (lines 776-781), and `VaultConfig` is itself derived from the owner and `vault_id` (lines 776-778, 783-785). To read the on-chain stream account, the provider must re-derive the vault PDA, which requires `vault_id` and `owner_public_key`. The spec makes this explicit: verification reads "that stream's on-chain account state for the session `vault_id`" (lines 672-674). The `StreamProof` only gives `stream_id`; the vault identity has to come from the session record, and it must be the *same* vault as the accepted proposal, not any vault the user controls.

4. `stream_id` pinning. After the first valid stream-proof, the provider pins the `stream_id`. This prevents the user from hopping between multiple on-chain streams that happen to match the params mid-session.

5. Deadline / liveness bookkeeping. "The first stream-proof-backed `ServiceRequest` MUST arrive by `create_stream_deadline`" (lines 668-669), and the session ends if that deadline passes without a first stream-proof (lines 690-692). The deadline is an off-chain/session concept (it is in `StreamParams`, not a stored on-chain stream field), so tracking whether the first stream-proof arrived in time is inherently session state.

### Why the protocol cannot simply be stateless

The premise "every `StreamProof` is accompanied by an on-chain state verification" is true, but on-chain verification only answers: *does an `ACTIVE` stream with these on-chain fields exist?* It does not answer:

- Was this stream opened under a proposal *I accepted* (with a session key *I recorded*)?
- Did it match the *accepted* `StreamParams` and the *pinned* policy, not just any current policy?
- Did the first proof arrive by the *accepted* deadline?
- Is the signature from the key the vault owner *authorized* for this session?

Those bindings are established in the proposal phase and are not re-stated on chain. So they must be remembered.

You *could* in principle push the context into every `StreamProof` — carry the session public key, the accepted `StreamParams`, the pinned policy, and the `VaultProof`/`owner_signature` that binds the key — and re-verify from scratch on each request. That is the "stateless" option. What it costs:

- Every request becomes as heavy as the initial proposal: full owner-signature verification, vault-proof checks, policy evaluation, instead of one symmetric-ish signature check over `request_data`.
- You still cannot avoid *some* state: the per-vault "MUST NOT send another `StreamProposal` to the same provider while a proposal is pending" rule (lines 633-634, 658) and the "cap parameter-adjustment attempts at 5 per vault within 600 seconds" recommendation (line 644-645) require a replay/attempt cache keyed by vault. A purely stateless node cannot enforce "one pending proposal per vault-provider pair," which is exactly the replay protection you flagged as the motivation.
- Putting the session public key on chain (so it could be read from `StreamConfig`) would remove one field of session state, but it is not in the current account layout and has privacy implications (session key visible to observers).

So the two-phase split is a deliberate state-vs-bandwidth trade: a heavy, state-creating proposal once, then lightweight per-request proofs that rely on the recorded session. The ongoing state is the price of lightweight proofs; the pending-proposal state is the price of per-vault replay protection.

### Direct answer to your sub-question

"Once the stream is created, do we need to track anything for the stream session off-chain?" Yes. Even after the on-chain stream exists, the provider still needs the session record (session key, accepted params, pinned policy, vault identity, pinned `stream_id`, and first-proof-by-deadline flag) to verify each `StreamProof`, because none of those bindings are encoded on the on-chain stream account and the `StreamProof` itself only carries `stream_id` + `signature`. The spec also explicitly allows this state to persist across on-chain pause/resume (lines 683-684), which only makes sense if the off-chain session is tracked independently of the on-chain lifecycle state.

The protocol could be made closer to stateless only by fattening every `StreamProof` to re-carry (and re-prove) the proposal context, and even then a small per-vault replay cache would remain. The current design chose minimal per-request proofs plus a server-side session record instead.

PR #305, opened 2026-03-31
Extended kad disco API update

SionoiS:
cc @jm-clius @tinniaru3005

### DIFF
--- a/docs/ift-ts/raw/extended-kad-disco.md
+++ b/docs/ift-ts/raw/extended-kad-disco.md
@@ -169,14 +169,19 @@ and advertising this node and its capabilities.
 In the case of Extended Kademlia Discovery,
 `stop()` will cancel the periodic task of [refreshing the propagated `XPR`](#record-propagation).
 
-### `start_advertising(const char* service_id)`
+### `start_advertising(const char* service_id, const byte* add)`
 
 Start advertising this node against any capability
-encoded as an input `service_id` string.
+encoded as an input `service_id` string
+and optionally an advertisement raw bytes.
 
 In the case of Extended Kademlia Discovery,
 `start_advertising()` will include the input `service_id`
 in the [regularly propagated `XPR`](#record-propagation).
+
+In the case of Service Discovery,
+`start_advertising()` CAN include in addition to `service_id`
+an already encoded advertisement in raw bytes form.
 
 ### `stop_advertising(const char* service_id)`
 
@@ -192,15 +197,18 @@ if it was previously included.
 
 Lookup and return records for peers supporting the capability encoded in the input `service_id` string,
 using the underlying discovery protocol.
-`service_id` is an OPTIONAL input argument.
-If unset, it indicates a lookup for peers supporting any (or zero) capabilities.
 
 In the case of Extended Kademlia Discovery,
+`service_id` MAY be included as an input argument.
+If unset, it indicates a lookup for peers supporting any (or zero) capabilities.
 `lookup()` will trigger the random walk [record discovery](#record-discovery),
 filtering discovered records based on `service_id`, if specified.
 If no `service_id` is specified,
 Extended Kademlia Discovery will just return a random selection of peer records,
 matching any capability.
+
+In the case of Service Discovery,
+`lookup()` MUST include a `service_id`.
 
 ## Copyright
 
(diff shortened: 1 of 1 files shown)


fbarbu15 (approved):

Cofson (commented):
Clean and focused update. 

1. **`CAN` → `MAY`**: "CAN include in addition to `service_id`" uses a non-standard keyword. Per RFC 2119, `MAY` is the correct term for optional behavior.

2. **Parameter name `add`**: The name is a bit ambiguous — could be read as a verb. Consider something more descriptive like `ad_data` or `additional_bytes`, or at minimum a brief inline description of what it represents.

docs/ift-ts/raw/extended-kad-disco.md
```diff
@@ -169,15 +169,20 @@ and advertising this node and its capabilities.
 In the case of Extended Kademlia Discovery,
 `stop()` will cancel the periodic task of [refreshing the propagated `XPR`](#record-propagation).
 
-### `start_advertising(const char* service_id)`
+### `start_advertising(const char* service_id, const byte* add)`
```
jm-clius:
```suggestion
### `start_advertising(const char* service_id, const byte* advertisement)`
```
Think it's better to be explicit here.

docs/ift-ts/raw/extended-kad-disco.md
```diff
@@ -169,15 +169,20 @@ and advertising this node and its capabilities.
...
 
 Start advertising this node against any capability
-encoded as an input `service_id` string.
+encoded as an input `service_id` string
+and optionally an advertisement raw bytes.
```
jm-clius, line 176:
```suggestion
and an `advertisement` raw bytes.
```
We can discuss, but as far as I can see it would add unnecessary complexity to make the `advertisement` optional, or do you mean that we might want the option to initialise the cap disc module with an advertisement and it will just use this cached one if no new `advertisement` is provided on the API?

docs/ift-ts/raw/extended-kad-disco.md
```diff
@@ -169,15 +169,20 @@ and advertising this node and its capabilities.
...
 
 In the case of Extended Kademlia Discovery,
 `start_advertising()` will include the input `service_id`
 in the [regularly propagated `XPR`](#record-propagation).
 
+In the case of Service Discovery,
+`start_advertising()` CAN include in addition to `service_id`
```
jm-clius, line 183: `CAN` is not an accepted RFC keyword (should be `MAY`), but as mentioned above I think it will simplify things if we make the `advertisement` required - we're already expecting the application to create and maintain this record and making it optional will make the API (more) stateful. I also don't think we have to differentiate between Extended Kademlia and Service Discovery. I'm hoping to move this API to it's own separate spec and it can be agnostic as to the underlying service discovery implementation.

docs/ift-ts/raw/extended-kad-disco.md
```diff
@@ -192,16 +197,19 @@ if it was previously included.
...
 In the case of Extended Kademlia Discovery,
+`service_id` MAY be included as an input argument.
+If unset, it indicates a lookup for peers supporting any (or zero) capabilities.
 `lookup()` will trigger the random walk [record discovery](#record-discovery),
 filtering discovered records based on `service_id`, if specified.
 If no `service_id` is specified,
 Extended Kademlia Discovery will just return a random selection of peer records,
 matching any capability.
 
+In the case of Service Discovery,
```
jm-clius, line 210: I don't think we should overfit our API with options specific to Extended Kademlia (in other words, it should focus on being exactly useful for Service Discovery by default and we can rather retrofit Extended Kademlia to fit the consistent API). As far as I remember we decided during the offsite to add two more API functions for discovery:
`start_discovering(const char* service_id)`
`stop_discovering(const char* service_id)`
to initialise or destruct the service discovery tables. In that case `lookup()` without service ID could perhaps still make sense for Service Discovery - at first we can just grab a random set of peers from each of the existing discovery tables, but eventually we can also consider implementing a simple random walk over the existing Kad-DHT (without service filtering, we wouldn't need XPRs published over the DHT).

jm-clius (commented):
Thanks! A couple of comments/questions below.

SionoiS:
What if we don't update this spec, discuss the API and create a different one instead? @jm-clius

SionoiS:
After discussion with @jm-clius we will close this.

In other PRs we will introduce a new spec for the API and remove it from this here spec.

PR #307, opened 2026-04-02
Add SURB specification (§8.7)

AkshayaMani:
This PR adds §8.7 Single-Use Reply Blocks (SURBs) to `mix.md`, specifying how SURBs are structured, created, used by recipients, processed by mix nodes, and recovered by the Exit Layer. Targets `chore/mix-restructure`.

#### Key changes

- **§8.7.1 SURB Component Sizes:** Defines the structure and byte sizes of SURBs 
- **§8.7.2 SURB Creation:** Specifies how the initiating node constructs a SURB and embeds it in the message, stores the reply key tuple for later recovery
- **§8.7.3 Using a SURB:** Specifies how the recipient wraps its reply in a Sphinx packet using the received SURB
- **§8.7.4 SURB Reply Processing:** Specifies how a mix node detects and processes an incoming SURB-based reply packet during Sphinx processing
- **§8.7.5 Reply Recovery:** Specifies how the initiating node recovers the plaintext reply by decrypting $L+1$ `AES-CTR` layers

#### Fixes and Cleanup

- Fix broken section cross-references
- Formatting fixes and typos

[!NOTE] §8.7.5 currently uses `AES-CTR` for reply decryption, consistent with the rest of the spec. The broader integrity limitation is tracked in #306 .

### DIFF
--- a/docs/ift-ts/raw/mix.md
+++ b/docs/ift-ts/raw/mix.md
@@ -359,7 +359,8 @@ Common strategies include Proof of Work (PoW), Verifiable Delay Functions (VDFs)
 The sender is responsible for appending the appropriate exit abuse prevention data (_e.g.,_ nonce, timestamp) to the message payload.
 The format and verification logic depend on the selected method.
 
-Note: The exit abuse prevention mechanisms described above are intended to protect the exit from flooding attacks routed through them.
+Note:
+The exit abuse prevention mechanisms described above are intended to protect the exit from flooding attacks routed through them.
 They do not provide protection against denial-of-service (DoS) or resource exhaustion attacks targeting the mixnet itself (_e.g.,_ flooding mix nodes with traffic, inducing processing overhead, or targeting bandwidth).
 
 Protections against attacks targeting the mixnet itself are addressed in [Section 6.6](#66-dos-protection).
@@ -634,6 +635,9 @@ The recommended total packet size is $4608$ bytes, chosen to:
 - Allow inclusion of additional data such as SURBs without requiring fragmentation,
 - Maintain reasonable per-hop processing and bandwidth overhead.
 
+Note:
+When [DoS protection](#66-dos-protection) is enabled, the maximum packet size MUST be increased to accommodate the proof size of the chosen DoS protection mechanism.
+
 This recommended total packet size of $4608$ bytes yields:
 
 $`
@@ -663,8 +667,6 @@ Implementers SHOULD align with widely used defaults to maximize anonymity set si
 Similarly, parameters such as $r$ and $t$ are configurable.
 Changes to these parameters affect header size and therefore impact payload size if the total packet size remains fixed.
 However, if such changes alter the total packet size on the wire, the same anonymity set considerations apply.
-
-When DoS protection is enabled (see [Section 6.6](#66-dos-protection)), the total wire size may increase depending on the integration approach.
 
 The following subsection defines how the next-hop or destination address and forwarding delay are encoded within $β$ to enable correct routing and mixing behavior.
 
--- a/docs/ift-ts/raw/mix.md
+++ b/docs/ift-ts/raw/mix.md
@@ -6,8 +6,8 @@
 | Slug | 99 |
 | Status | raw |
 | Category | Standards Track |
-| Editor | Prem Prathi <prem@status.im>|
-| Contributors | Akshaya Mani <akshaya@status.im>, Daniel Kaiser <danielkaiser@status.im>, Hanno Cornelius |
+| Editor | Akshaya Mani <akshaya@status.im>|
+| Contributors | Prem Prathi <prem@status.im>, Daniel Kaiser <danielkaiser@status.im>, Hanno Cornelius |
 
 <!-- timeline:start -->
 
@@ -116,7 +116,7 @@ However, Tor relies on session persistence and is ill-suited for protocols requi
 
 The Mix Protocol addresses this gap with a decentralized message routing layer based on classical mix network principles.
 It applies layered encryption and per-hop delays to obscure both routing paths and timing correlations.
-Each message is routed independently, providing resistance to traffic analysis and protection against metadata leakage
+Each message is routed independently, providing resistance to traffic analysis and protection against metadata leakage.
 
 By decoupling anonymity from connection state and transport negotiation, the Mix Protocol offers a modular privacy abstraction that existing libp2p protocols can adopt without altering their core behavior.
 
@@ -177,7 +177,7 @@ Even mix nodes learn only the immediate routing information and the delay to be 
 They do not learn their position in the path or the total number of hops.
 
 The packet format is resistant to tagging and replay attacks and is compact and efficient to process.
-Sphinx packets also include per-hop integrity checks and enforces a maximum path length.
+Sphinx packets also include per-hop integrity checks and enforce a maximum path length.
 Together with a constant-size header and payload, this provides bounded protection against endless routing and malformed packet propagation.
 
 It also supports anonymous and indistinguishable reply messages through [Single-Use Reply Blocks (SURBs)](https://cypherpunks.ca/~iang/pubs/Sphinx_Oakland09.pdf), although reply support is not implemented yet.
@@ -297,8 +297,8 @@ Mix Protocol implementations MUST NOT assume persistent stream availability and 
 Pluggable components define functionality that extends or configures the behavior of the Mix Protocol beyond its core message routing logic.
 Each component in this section falls into one of two categories:
 
-- Required for interoperability and path construction (_e.g.,_ discovery, delay strategy).
-- Optional or deployment-specific (_e.g._ spam protection, cover traffic, incentivization).
+- Required for interoperability and path construction (_e.g.,_ discovery, delay strategy, DoS protection).
+- Optional or deployment-specific (_e.g.,_ exit abuse prevention, cover traffic, incentivization).
 
 The following subsections describe the role and expected behavior of each.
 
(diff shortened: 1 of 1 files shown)


AkshayaMani:
@bkomuves Leaving this here as discussed. Feel free to build on it or flag anything that conflicts with your Sphinx write-up.

@mghazwi

Cofson (commented):
Very thorough SURB specification, great work!

In mixed capitalization when referencing steps — some places use uppercase "Step" and others lowercase "step", sometimes within the same sentence (_e.g.,_ "Step 2 and step 3.a" in §8.7.2). The existing spec sections that this PR touches are also changed in both directions (some "Step" → "step", some "step" → "Step").

As a nitpicky suggestion: Pick one convention and apply it consistently. A common approach: capitalize when used as a label reference ("see Step 1"), lowercase when inline in prose ("in step 3.c."). Either way, it should be uniform throughout.

AkshayaMani:
> In mixed capitalization when referencing steps — some places use uppercase "Step" and others lowercase "step", sometimes within the same sentence (_e.g.,_ "Step 2 and step 3.a" in §8.7.2). The existing spec sections that this PR touches are also changed in both directions (some "Step" → "step", some "step" → "Step").
> 
> As a nitpicky suggestion: Pick one convention and apply it consistently. A common approach: capitalize when used as a label reference ("see Step 1"), lowercase when inline in prose ("in step 3.c."). Either way, it should be uniform throughout.

The current state already has an implicit convention:
- Top-level step (like 1., 2.) → uppercase
- Sub-step (like 3.a, 3.b) → lowercase

But you're right that it looks inconsistent when both appear in the same sentence (line 1274: "Step 2 and step 3.a"). Implemented the suggestion.

AkshayaMani:
@bkomuves @mghazwi Thinking of merging this to keep it contained and for completeness. The message integrity issue (#306 ) and any restructuring of the spec can be addressed in follow-up PRs. Let me know if you have any concerns.

mghazwi:
> @bkomuves @mghazwi Thinking of merging this to keep it contained and for completeness. The message integrity issue (#306 ) and any restructuring of the spec can be addressed in follow-up PRs. Let me know if you have any concerns.

Yes sure, this can be merged for now, and we can modify/improve/restructure later in future PRs. Possibly taking out the sphinx and SURBs in a separate spec.

I have few comments as I read the SURBs part but these are not critical.

docs/ift-ts/raw/mix.md
mghazwi, line 849: I think it would be much cleaner and more consistent if this part also shows where the SURBs id fits in the header so this is basically:
$Δ \mid \mathrm{delay}_i \mid id \mid 0_{((t+1)(r-L)+1)κ$

which is to say:
- if there is an address Δ and delay (both non-zero) then relay to Δ
- if there is an address Δ but delay is zero then it is an exit, and will forward to Δ
- if Δ and delay are zeros then it is a SURB meant for the current node's exit layer.

I guess we are assuming here that SURBs final destination is the exit layer of the mix node that created the SURB, which is not the case for forward message since those always have Δ.

docs/ift-ts/raw/mix.md
```diff
@@ -1210,9 +1216,214 @@ Once the node determines its role as an exit following the steps in [Section 8.6
...
+   - Hand off the serialized application message, the origin protocol codec, destination address $Δ$ (extracted in step 1.), and any SURBs extracted in step 3. to the local Exit Layer for further processing and delivery.
 
    - The Exit Layer is responsible for establishing a client-only connection and forwarding the message to the destination. Implementations MAY reuse an existing stream to the destination, if doing so does not introduce any observable linkability between forwarded messages.
+     It is also responsible for storing any received SURBs and routing responses from the destination using them (see [Section 8.7.3](#873-using-a-surb)).
+
+### 8.7 Single-Use Reply Blocks
+
+A Single-Use Reply Block (SURB) allows the recipient of a Sphinx packet to send a reply without learning the sender's identity, the return mix path, or any forwarding delays.
+
+A SURB MUST be used at most once. Reusing a SURB allows nodes on the return path to link multiple replies back to the original sender.
```
mghazwi: So this line doesn't mention how this is achieved? what makes them single-use? I assume the way this is enforced here is tagging and replay prevention? unless I overlooked it somewhere, might be better to mention this.

docs/ift-ts/raw/mix.md
```diff
@@ -1210,9 +1216,214 @@ Once the node determines its role as an exit following the steps in [Section 8.6
...
+
+3. **Construct Return Path Header**
+
+   Compute per-hop filler strings and construct the routing header following the same procedure in [Section 8.5.2](#852-construction-steps) Steps 3.b and 3.c, with just $β_{L-1}$ computed as follows:
+
+   $`
+   \begin{array}{l}
+   β_{L-1} = \mathrm{AES\text{-}CTR}\bigl(β_{\mathrm{aes\_key}_{L-1}},
+   β_{\mathrm{iv}_{L-1}},
+   0_{tκ} \mid \mathrm{id} \mid 0_{(((t+1)(r-L))+1)κ}
```
mghazwi: same comment as above, I think for clarity/consistency I would place the same components here so this would be:
$ Δ \mid \mathrm{delay}i \mid id \mid 0{((t+1)(r-L)+1)κ $
where in here both Δ and delay would be 0.

mghazwi (commented):
looks good, just few small comments.

docs/ift-ts/raw/mix.md
```diff
@@ -1210,9 +1216,214 @@ Once the node determines its role as an exit following the steps in [Section 8.6
...
    - The Exit Layer is responsible for establishing a client-only connection and forwarding the message to the destination. Implementations MAY reuse an existing stream to the destination, if doing so does not introduce any observable linkability between forwarded messages.
+     It is also responsible for storing any received SURBs and routing responses from the destination using them (see [Section 8.7.3](#873-using-a-surb)).
+
+### 8.7 Single-Use Reply Blocks
+
+A Single-Use Reply Block (SURB) allows the recipient of a Sphinx packet to send a reply without learning the sender's identity, the return mix path, or any forwarding delays.
+
+A SURB MUST be used at most once. Reusing a SURB allows nodes on the return path to link multiple replies back to the original sender.
+
+A SURB encodes a complete Sphinx header for a return path, a symmetric reply key, and a unique reply identifier. The initiating node constructs one or more SURBs and embeds them in the outgoing Sphinx packet payload. The recipient uses a SURB to reply — only the original sender can decrypt the reply.
```
jm-clius, line 1232: Minor note: it's a good idea to ease reviews/maintenance to use [semantic line breaks](https://sembr.org/) throughout. Single line breaks won't actually be rendered in markdown, but makes it easier to review line-by-line.

For example:

```suggestion
A SURB encodes a complete Sphinx header for a return path,
a symmetric reply key,
and a unique reply identifier.
The initiating node constructs one or more SURBs
and embeds them in the outgoing Sphinx packet payload.
The recipient uses a SURB to reply — only the original sender can decrypt the reply.
```
The above will still render as a single paragraph on the website, but now it's easier to review each semantic idea separately in a separate line.

docs/ift-ts/raw/mix.md
jm-clius: Same question. If I understand correctly, though, it's just the sender's responsibility here not to reuse the same SURB for multiple requests? One way to avoid this ambiguity is to avoid passive voice (often good idea in specs) and make it clear who the agent is, e.g. `The sender MUST NOT use the same SURB more than once.` This shows that it's not enforced but prescribed.

jm-clius (approved):
Unless there're major objections to the content (I don't have the background to find any 😅) I'd suggest merging this. Minor changes, clarifications, etc. can be added in subsequent PRs.

chaitanyaprem (approved):
LGTM 
Thanks! I understand SURB better now at some level. 
Will re-read sphinx packet format to get deeper understanding and then maybe i will get more indepth understanding.

Cofson (approved):

chaitanyaprem:
@AkshayaMani can we resolve conflicts and merge this? We can take enhancements as separate PR.

docs/ift-ts/raw/mix.md
AkshayaMani: §8.5.2 forward exit formula describes only the forward Sphinx construction, where the exit block always contains a destination address $Δ$. There is no $id$ field in a forward exit block. The SURB exit construction is a separate variant defined in §8.7.2.

> I guess we are assuming here that SURBs final destination is the exit layer of the mix node that created the SURB, which is not the case for forward message since those always have $Δ$.

Yes, SURB replies are always destined for the initiating node's Exit Layer.

docs/ift-ts/raw/mix.md
AkshayaMani: > e.g. The sender MUST NOT use the same SURB more than once. This shows that it's not enforced but prescribed.

Changed to:  The recipient MUST NOT use the same SURB more than once.

> So this line doesn't mention how this is achieved? what makes them single-use? I assume the way this is enforced here is tagging and replay prevention?

Clarified here and also added a `Note` in §8.6.1.

docs/ift-ts/raw/mix.md
AkshayaMani: Good suggestion. Updated the formula to split the zero address and zero delay into separate blocks, making the structure visually parallel to the forward exit formula in §8.5.2

docs/ift-ts/raw/mix.md
AkshayaMani: Good point. Adopting semantic line breaks would touch the entire file though.'ll defer it to a separate stylistic cleanup PR than mixed in with this one.

PR #309, opened 2026-04-07
New logos service discovery API

SionoiS:
Extracted from https://github.com/vacp2p/rfc-index/blob/master/docs/ift-ts/raw/extended-kad-disco.md#api-specification

This spec describe the future logos service discovery API.

cc @jm-clius @tinniaru3005

### DIFF
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -75,7 +75,7 @@ for the service encoded in the input `service_id` string.
 
 ### `RawAdvertisement* lookup(const char* service_id)`
 
-Lookup and return records for peers supporting
+Lookup and return advertisements for peers supporting
 the service encoded in the input `service_id` string,
 using the underlying discovery protocol.
 
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -52,9 +52,11 @@ of the underlying discovery protocol (such as advertising or discovery loops).
 
 ### `start_advertising(const char* service_id, const byte* advertisement)`
 
-Start advertising this node against the service
-encoded as an input `service_id` string
-and an `advertisement` raw bytes.
+Start advertising the encoded raw `advertisement`
+against any service encoded as a `service_id` string.
+For peer discovery,
+the node MUST encode sufficient connection information in the `advertisement`
+for discoverers to connect to it.
 
 ### `stop_advertising(const char* service_id)`
 
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -60,15 +60,29 @@ and an `advertisement` raw bytes.
 Stop advertising this node against the service
 encoded in the input `service_id` string.
 
+### `start_discovering(const char* service_id)`
+
+Start discovering and maintaining search tables
+for the service encoded in the input `service_id` string.
+
+### `stop_discovering(const char* service_id)`
+
+Stop discovering and maintenance of search tables
+for the service encoded in the input `service_id` string.
+
 ### `ExtensiblePeerRecords* lookup(const char* service_id)`
 
 Lookup and return records for peers supporting
 the service encoded in the input `service_id` string,
 using the underlying discovery protocol.
 
+It is RECOMMENDED to use `start_discovering`
+in advance of any `lookup` for each `service_id`
+as a way to speed up search.
+
 ### `ExtensiblePeerRecords* lookup_random()`
 
-Reserved for future use.
+Unused, reserved for future use.
 
 ## Copyright
 
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -83,7 +83,7 @@ It is RECOMMENDED to use `start_discovering`
 in advance of any `lookup` for each `service_id`
 as a way to speed up search.
 
-### `ExtensiblePeerRecords* lookup_random()`
+### `RawAdvertisement* lookup_random()`
 
 Unused, reserved for future use.
 
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -47,8 +47,8 @@ of the underlying discovery protocol (such as initialising the routing table).
 ### `stop()`
 
 Stop the discovery protocol,
-including all tasks related to node maintenance
-and advertising of its services.
+including all tasks related to maintenance
+of the underlying discovery protocol (such as advertising or discovery loops).
 
 ### `start_advertising(const char* service_id, const byte* advertisement)`
 
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -41,8 +41,8 @@ all functions SHOULD return an error result type appropriate to the implementati
 ### `start()`
 
 Start the discovery protocol,
-including all tasks related to bootstrapping, maintenance
-and advertising of this node and its services.
+including all tasks related to bootstrapping and maintenance
+of the underlying discovery protocol (such as initialising the routing table).
 
 ### `stop()`
 
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -28,7 +28,8 @@ and [extensible peer records specification](https://github.com/vacp2p/rfc-index/
 
 ## API Specification
 
-The aim is to define an API that is compatible with most discovery protocols,
+The aim is to define an API that is compatible with most discovery protocols
+that supports service-specific discovery,
 maintaining similar function signatures even if the underlying protocol differs.
 
 The API is defined in the form of C-style bindings.
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -73,7 +73,7 @@ for the service encoded in the input `service_id` string.
 Stop discovering and maintenance of search tables
 for the service encoded in the input `service_id` string.
 
-### `ExtensiblePeerRecords* lookup(const char* service_id)`
+### `RawAdvertisement* lookup(const char* service_id)`
 
 Lookup and return records for peers supporting
 the service encoded in the input `service_id` string,
--- a/docs/ift-ts/raw/logos-discovery-api.md
+++ b/docs/ift-ts/raw/logos-discovery-api.md
@@ -1,0 +1,83 @@
+# LOGOS-DISCOVERY-API
+
+| Field | Value |
+| --- | --- |
+| Name | Logos Discovery API |
+| Slug | 145  |
+| Status | raw |
+| Category | Standards Track |
+| Editor | Simon-Pierre Vivier <simvivier@status.im> |
+| Contributors | Hanno Cornelius <hanno@status.im>|
+
+## Abstract
+
+TODO
+
+## Motivation
+
+TODO
+
+## Semantic
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”,
+“SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document
+are to be interpreted as described in [2119](https://www.ietf.org/rfc/rfc2119.txt).
+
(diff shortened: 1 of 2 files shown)


docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
+# LOGOS-DISCOVERY-API
+
+| Field | Value |
+| --- | --- |
+| Name | Logos Discovery API |
```
jm-clius: Perhaps something like `Service Discovery API` or `Kademlia Service Discovery API`? I think we want to remove `Logos` from the names here, as we're designing an API for lower-level libp2p protocols

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”,
+“SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document
+are to be interpreted as described in [2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+Please refer to [libp2p Kademlia DHT specification](https://github.com/libp2p/specs/blob/e87cb1c32a666c2229d3b9bb8f9ce1d9cfdaa8a9/kad-dht/README.md) (`Kad-DHT`)
+and [extensible peer records specification](https://github.com/vacp2p/rfc-index/blob/513d8eae6be8b7b30bf427023ac686df2f2918c0/docs/ift-ts/raw/extensible-peer-records.md) (`XPR`) for terminology used in this document.
+
+## API Specification
+
+The aim is to define an API that is compatible with most discovery protocols,
```
jm-clius:
```suggestion
The aim is to define an API that is compatible with most discovery protocols
that supports service-specific discovery,
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+However, this simply serves to illustrate the exposed functions
+and can be adapted into the conventions of any strongly typed language.
+Although unspecified in the API below,
+all functions SHOULD return an error result type appropriate to the implementation language.
+
+### `start()`
+
+Start the discovery protocol,
+including all tasks related to bootstrapping, maintenance
+and advertising of this node and its services.
```
jm-clius: I think the subtlety here is that there is no longer a "this node" that advertises implied by the API - just a protocol that can advertise any `advertisements`.

```suggestion
Start the discovery protocol,
including all tasks related to bootstrapping and maintenance
of the underlying discovery protocol (such as initialising the routing table).
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+
+### `stop()`
+
+Stop the discovery protocol,
+including all tasks related to node maintenance
+and advertising of its services.
```
jm-clius:
```suggestion
Stop the discovery protocol,
including all tasks related to maintenance
of the underlying discovery protocol (such as advertising or discovery loops).
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+
+### `start_advertising(const char* service_id, const byte* advertisement)`
+
+Start advertising this node against the service
+encoded as an input `service_id` string
+and an `advertisement` raw bytes.
```
jm-clius:
```suggestion
Start advertising the encoded raw `advertisement`
against any service encoded as a `service_id` string.
For peer discovery,
the node MUST encode sufficient connection information in the `advertisement`
for discoverers to connect to it.
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+including all tasks related to bootstrapping, maintenance
+and advertising of this node and its services.
+
+### `stop()`
+
+Stop the discovery protocol,
+including all tasks related to node maintenance
+and advertising of its services.
+
+### `start_advertising(const char* service_id, const byte* advertisement)`
```
jm-clius: I know we're not necessarily trying to write correct C in our API definition, but I don't think `byte*` exist as a type in C. I'd suggest something like:

```suggestion
### `start_advertising(const char* service_id, const RawAdvertisement advertisement)`
```

And then just preface this section with a `struct`:

```
typedef struct {
    uint8_t* advertisementBytes;
    size_t   length;
} RawAdvertisement;
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+
+Start discovering and maintaining search tables
+for the service encoded in the input `service_id` string.
+
+### `stop_discovering(const char* service_id)`
+
+Stop discovering and maintenance of search tables
+for the service encoded in the input `service_id` string.
+
+### `ExtensiblePeerRecords* lookup(const char* service_id)`
```
jm-clius:
```suggestion
### `RawAdvertisement* lookup(const char* service_id)`
```
If we define the `RawAdvertisement` struct elsewhere.

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+
+Lookup and return records for peers supporting
+the service encoded in the input `service_id` string,
+using the underlying discovery protocol.
+
+It is RECOMMENDED to use `start_discovering`
+in advance of any `lookup` for each `service_id`
+as a way to speed up search.
+
+### `ExtensiblePeerRecords* lookup_random()`
```
jm-clius:
```suggestion
### `RawAdvertisement* lookup_random()`
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+for the service encoded in the input `service_id` string.
+
+### `stop_discovering(const char* service_id)`
+
+Stop discovering and maintenance of search tables
+for the service encoded in the input `service_id` string.
+
+### `ExtensiblePeerRecords* lookup(const char* service_id)`
+
+Lookup and return records for peers supporting
```
jm-clius:
```suggestion
Lookup and return advertisements for peers supporting
```

docs/ift-ts/raw/logos-discovery-api.md
```diff
@@ -0,0 +1,97 @@
...
+TODO
+
+## Semantic
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”,
+“SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document
+are to be interpreted as described in [2119](https://www.ietf.org/rfc/rfc2119.txt).
+
+Please refer to [libp2p Kademlia DHT specification](https://github.com/libp2p/specs/blob/e87cb1c32a666c2229d3b9bb8f9ce1d9cfdaa8a9/kad-dht/README.md) (`Kad-DHT`)
+and [extensible peer records specification](https://github.com/vacp2p/rfc-index/blob/513d8eae6be8b7b30bf427023ac686df2f2918c0/docs/ift-ts/raw/extensible-peer-records.md) (`XPR`) for terminology used in this document.
```
jm-clius: I think we probably can avoid having to reference XPR here by making the returned records for `lookup()` also raw advertisement bytes. Think this will be easier to implement in actual C bindings too. More below.

jm-clius (changes requested):
Thanks! Added some suggestions.

jm-clius (approved):
Thank! Actually, I think this can be marked as ready for review and merged. Will be good to share with p2p team as well.

SionoiS:
FYI @richard-ramos @gmelodie @vladopajic

This will be the discovery logos module API

gmelodie (commented):
Question: do you folks think it'd be a good idea to have a `discover(service)` proc that would be more opinionated but a bit easier to use? Like it could start discovering for 5s and then return whatever it finds or, if discovery is already started, return what it knows about immediately. Don't know if this should be on this spec though, just brainstorming

SionoiS:
> Like it could start discovering for 5s and then return whatever it finds

Yes that would be `random_lookup` main functionality but we don't need that just yet.

> if discovery is already started, return what it knows about immediately

"immediately" does a lot of work here. There will always be a request send to nodes to request adverts.

We can for sure return `PeerId`s and addresses for nodes in the search table but the API is really about adverts not connection information directly.

gmelodie:
@SionoiS makes sense!

gmelodie (approved):

docs/ift-ts/raw/service-discovery-api.md
```diff
@@ -0,0 +1,109 @@
...
+| Contributors | Hanno Cornelius <hanno@status.im>|
+
+## Abstract
+
+TODO
+
+## Motivation
+
+TODO
+
```
gmelodie, line 19: Not sure what the protocol is to merge specs, but I'd expect there to be at least a rough attempt to fill these fields

docs/ift-ts/raw/service-discovery-api.md
```diff
@@ -0,0 +1,109 @@
...
+typedef struct {
+    const uint8_t* advertisementBytes;
+    size_t   length;
+} RawAdvertisement;
+
+typedef struct {
+    const RawAdvertisement* items;
+    size_t                  count;
+} RawAdvertisementList;
```
gmelodie: - Any reason to use `Raw` on names?
- I like smaller names better
- nits: `length -> len`; `count -> len`
```suggestion
typedef struct {
    const uint8_t* bytes;
    size_t   len;
} Advertisement;

typedef struct {
    const Advertisement* ads;
    size_t                  len;
} AdvertisementList;
```

gmelodie (approved):

Cofson (commented):
Thanks, looks good. Couple of things below.

docs/ift-ts/raw/service-discovery-api.md
```diff
@@ -0,0 +1,109 @@
...
+including all tasks related to bootstrapping and maintenance
+of the underlying discovery protocol (such as initialising the routing table).
+
+### `stop()`
+
+Stop the discovery protocol,
+including all tasks related to maintenance
+of the underlying discovery protocol (such as advertising or discovery loops).
+
+### `start_advertising(const char* service_id, const RawAdvertisement* advertisement)`
```
Cofson: The type defs above use `Advertisement` / `AdvertisementList`, but the function signatures still say `RawAdvertisement` / `RawAdvertisementList`. Looks like the rename only landed in one place.

docs/ift-ts/raw/service-discovery-api.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Motivation
+
+TODO
+
+## Semantic
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”,
+“SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document
+are to be interpreted as described in [2119](https://www.ietf.org/rfc/rfc2119.txt).
```
Cofson:
```suggestion
are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
```

docs/ift-ts/raw/service-discovery-api.md
```diff
@@ -0,0 +1,109 @@
...
+
+Unused, reserved for future use.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+## References
+
+- [logos service discovery](https://github.com/vacp2p/rfc-index/blob/155c310d7bfad6ea3cd9f68e45c68dad731ff629/docs/ift-ts/raw/logos-service-discovery.md)
```
Cofson: Repo moved to `logos-co/logos-lips`, better to point there directly.

```suggestion
- [logos service discovery](https://github.com/logos-co/logos-lips/blob/155c310d7bfad6ea3cd9f68e45c68dad731ff629/docs/ift-ts/raw/logos-service-discovery.md)
```

jm-clius:
@SionoiS anything outstanding to get this merged?

SionoiS:
> @SionoiS anything outstanding to get this merged?

We need a :heavy_check_mark: from a maintainer?

jm-clius:
@Cofson any further blockers? :)

fbarbu15 (approved):
LGTM, thanks

PR #311, opened 2026-04-08
feat: Mix Cover Traffic specification

chaitanyaprem:
## Summary

This PR introduces the Mix Cover Traffic specification — a concrete instantiation of the pluggable cover traffic component defined in Mix Protocol §6.4. The spec defines how mix nodes generate and emit cover packets to achieve sender unobservability.

### Key design decisions

- **Self-balancing R-slot pool**: Cover, locally originated messages, and forwarded packets share a single per-epoch budget `R` from the DoS protection mechanism. Forwarding naturally consumes ~`L/(1+L)` of the budget; the remainder is used for cover. No explicit origination rate constraint is needed.
- **Self-exit loop paths**: Cover packets use loop paths where the originator is the exit node. A dedicated codec string `/mix/cover-traffic/1.0.0 `identifies returning cover packets during exit processing, handled locally.
- **Two strategies**: Constant-Rate (RECOMMENDED default) and Poisson-Rate (for Loopix/Nym alignment). The spec includes detailed analysis of why Loopix formal guarantees don't transfer unchanged to this architecture (shared budget violates independence, epoch boundaries break memorylessness).
- **Pre-computation as optional optimization**: Cover packets MAY be pre-built during epoch `N-1` for epoch `N`, with batched pre-computation as an implementation recommendation.
- **Per-hop proof assumption** (§4): The budget model assumes per-hop generated proofs. Sender-generated proofs require a different budget model, deferred to future work.
- **Pre-scheduled emission timing** (§11.3): A future enhancement where all traffic types share a fixed timing grid, eliminating periodic pattern concerns. Only compatible with Constant-Rate. Inspired by blend protocol.

### Changes to existing specs

- **mix.md §8.6.4**: Added step 4 "Internal Protocol Codec Check" between extracting the codec and Exit Layer handoff. This is a generic hook for pluggable components that use internal protocol codecs.
- **mix-dos-protection.md §8.2.3**: Added `OnEpochChange(callback)` procedure for epoch boundary notifications to pluggable components.

### DIFF
--- a/docs/ift-ts/raw/mix-cover-traffic.md
+++ b/docs/ift-ts/raw/mix-cover-traffic.md
@@ -1,0 +1,712 @@
+# Mix Cover Traffic
+
+| Field | Value |
+| --- | --- |
+| Name | Mix Cover Traffic |
+| Slug |  |
+| Status | raw |
+| Category | Standards Track |
+| Editor | Prem Prathi <prem@status.im> |
+| Contributors |  |
+
+<!-- timeline:start -->
+
+## Timeline
+
+<!-- timeline:end -->
+
+## Abstract
+
+This document specifies the cover traffic architecture for the [libp2p Mix Protocol](./mix.md).
+The architecture ensures that an observer cannot distinguish cover traffic from locally originated messages
+by observing a node's emission pattern.
+It defines how cover packets are generated and emitted,
+how the rate-limit budget is shared across cover and non-cover traffic,
+and specifies the Constant-Rate cover traffic strategy, with Poisson-Rate as a future consideration in §11.5.
+
+## 1. Introduction
+
+The Mix Protocol provides sender anonymity through layered encryption and per-hop delays.
+However, without cover traffic,
+an adversary observing a mix node's emission rate can mount several attacks:
+
+- **Traffic analysis**: by correlating emission bursts with known events,
+  an adversary can link a node's activity periods to specific senders or recipients.
+- **Intersection attack**: by observing which nodes are active each time a message reaches its destination,
+  an adversary can progressively narrow down the set of possible senders across multiple messages.
+- **Timing correlation**: by matching idle and active periods across mix nodes,
+  an adversary can correlate ingress and egress packets.
+
+All three attacks rely on the same weakness:
+a node's emission pattern leaks information about whether it is carrying non-cover traffic.
+
+Cover traffic addresses this by ensuring a node's emission pattern does not depend on non-cover traffic volume,
+making it indistinguishable to an observer whether the node is sending locally originated messages or none at all.
+
+The Mix Protocol defines cover traffic as a pluggable component (see [Mix Protocol §6.4](./mix.md#64-cover-traffic)).
+This specification provides a concrete instantiation of that component,
+defining the cover traffic architecture, the rate-limit budget model, and two concrete scheduling strategies.
+The architecture is designed to be compatible with the DoS protection mechanism defined in [Mix DoS Protection](./mix-dos-protection.md)
+and specifically with the [Mix RLN DoS Protection](./mix-spam-protection-rln.md) mechanism.
+
+## 2. Terminology
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL"
+in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+Other terms used in this document are as defined in the [libp2p Mix Protocol](./mix.md) and [Mix DoS Protection](./mix-dos-protection.md).
+
+The following additional terms are used throughout this specification:
+
+- **Cover Packet**
+  A dummy Sphinx packet that carries no application payload
+  and is indistinguishable from non-cover Sphinx packets in structure, size, and routing behavior.
+
+- **Slot**
+  A single rate-limit token within an epoch's budget of `R` tokens, as defined by the DoS protection mechanism.
+  Each outgoing packet — whether cover or non-cover — consumes exactly one slot.
+
+- **Slot Pool**
+  The collection of rate-limit slots available for a given epoch.
+
+- **Epoch**
+  A fixed time window of duration `P` seconds during which each mix node is permitted to emit at most `R` packets,
+  as enforced by the DoS protection mechanism.
+
+## 3. Design Principles
+
+The cover traffic architecture is guided by the following principles:
+
+- **Sender unobservability**: A node's emission pattern must not depend on non-cover traffic volume,
+  making it indistinguishable to an observer whether the node is carrying non-cover traffic or not.
+- **Indistinguishability**: Cover packets are structurally identical to non-cover Sphinx packets in size and routing behavior,
+  preventing packet-level classification. ([Mix Protocol §6.4](./mix.md#64-cover-traffic))
+- **Self-exit**: Cover packets SHOULD use loop paths where the originating node is also the exit node.
+  This ensures the dummy payload is never decrypted by an external party,
+  eliminating the risk of cover classification at the exit.
+- **DoS protection compliance**: All cover traffic operates within the rate-limit budget `R` enforced per epoch.
+  Proofs are epoch-bound and unused slots are discarded at epoch boundaries. ([Mix DoS Protection](./mix-dos-protection.md))
+- **Slot integrity**: Each rate-limit slot is consumed at most once on the wire per epoch.
+  When a non-cover claim reclaims a slot held by a queued cover packet,
+  that packet's pre-computed proof is discarded before the slot is reused.
+- **Mix node only**: Cover traffic is generated only for mix nodes that act as intermediate nodes forwarding mix traffic
+  and participate continuously in the network.
+  Initiating-only nodes are mostly short-lived with dynamic identifiers and do not forward traffic,
+  making cover traffic neither practical nor beneficial for them.
+- **Pre-computation**: As an optimization, cover packets and their proofs MAY be generated during epoch `N-1`,
+  so they are ready to emit at the start of epoch `N` without any cryptographic work at emission time.
+
+## Overview
+
+A mix node plays multiple roles at once: it sends its own messages, relays messages for other nodes, and ideally hides which of these it is doing.
+Without protection, an observer watching the node's outgoing packets can tell when it is active, how busy it is, and when it is idle —
+enough to link users to their messages through traffic patterns.
+
+Cover traffic addresses this by emitting additional dummy packets that look identical to real mix traffic from the outside.
+An observer still sees packets leaving the node,
+but can no longer tell from the pattern alone whether those packets are real or dummy.
+
+The node operates under a rate limit that bounds total packets emitted per epoch ([§4](#4-rate-limit-budget-model)).
+Every packet — cover, locally originated, or forwarded — consumes one slot from this shared budget.
+Forwarding typically takes a large share of the budget because each originated packet traverses multiple hops,
+so the maximum cover rate is naturally bounded below the total.
+
+Cover is emitted at a steady configurable rate, up to this bound ([§7.1](#71-constant-rate-cover-traffic)).
+A `cover_rate_fraction` parameter scales cover down from the maximum,
+leaving headroom in the budget for spikes in real traffic.
+Real traffic (locally originated and forwarded) claims slots from the same pool as it arrives;
+cover yields whatever slots remain.
+
+Cover packets follow round-trip paths — the sender is also the final destination,
+so the dummy payload is never decrypted by another party ([§5.1](#51-cover-packet-transmission)).
+For efficiency, cover packets and their rate-limit proofs MAY be pre-built during the previous epoch ([§6.1](#61-at-epoch-boundary))
+and revalidated at send time in case the underlying state has changed ([§6.5](#65-pre-computed-proof-validation-at-send-time)).
+
+Each epoch begins by discarding previous state and initializing a fresh slot budget,
+loading any pre-built cover packets prepared during the prior epoch.
+Throughout the epoch, cover, locally originated, and forwarded packets independently claim slots.
+Near the midpoint, the node starts pre-computing cover packets for the next epoch.
+At epoch end, unused slots are discarded and the cycle repeats.
+
+The specification focuses on the Constant-Rate strategy.
+An alternative Poisson-Rate strategy, where cover emission times are randomized, is kept for future consideration in [§11.5](#115-poisson-rate-cover-traffic).
+
+## 4. Rate Limit Budget Model
+
+Each mix node receives a budget of `R` slots per epoch from the DoS protection mechanism.
+Cover emission, locally originated message sending, and packet forwarding all draw from the same pool.
+Since each originated packet traverses `L` forwarding hops — where `L` is the mix path length
+as defined in [Mix Protocol §6](./mix.md#6-pluggable-components) —
+forwarding traffic naturally consumes a significant portion of the budget.
+
+If every node originates at rate `C` packets per epoch (cover plus locally originated combined),
+each node forwards approximately `C * L` packets per epoch.
+Since origination and forwarding share the same budget `R`:
+
+```
+C + C * L ≤ R
+C ≤ R / (1 + L)
+```
+
+`R / (1 + L)` is therefore the **upper bound on total origination**,
+not a target for cover emission alone.
+Cover rate does not need to be explicitly reduced by a node's locally originated rate,
+because the slot pool is self-balancing (see below).
+
+This means the actual cover traffic emitted by a node is always less than `R` and depends on:
+
+- **Path length `L`**: longer paths consume more forwarding slots, leaving fewer for cover.
+  For `L=3`, approximately 25% of slots are available for cover and locally originated message sending.
+- **Network size `N` and forwarding variance**: with random path selection, forwarding load is not uniform.
+  Some nodes receive more forwarding traffic than the equilibrium average, leaving even fewer slots for cover.
+  The actual cover output per node therefore varies with network conditions.
+
+The slot pool is self-balancing — no explicit origination rate constraint is needed.
+Heavier forwarding load automatically leaves fewer slots for cover; lighter load leaves more.
+
+**Note on DoS protection architecture:**
+The self-balancing pool model assumes per-hop generated proofs
+([Mix DoS Protection §4.2](./mix-dos-protection.md#42-per-hop-generated-proofs)),
+where forwarding consumes slots from the node's own `R` budget.
+With sender-generated proofs ([Mix DoS Protection §4.1](./mix-dos-protection.md#41-sender-generated-proofs)),
+forwarding nodes only verify proofs and do not consume their own `R`,
+but cover emission must still account for forwarding load to maintain constant total output.
+The budget model and slot pool semantics for sender-generated proofs require separate analysis
+and are deferred to [§11.4](#114-budget-model-for-sender-generated-proofs).
+
+## 5. Integration with the Mix Protocol
+
+The cover traffic mechanism integrates with the Mix Protocol at four points in packet processing.
+
+Cover packets are identified by the reserved protocol codec `"/mix/cover/1.0.0"`.
+This codec is used as the origin protocol codec during Sphinx packet construction
+and is checked during exit processing to distinguish cover packets from application traffic.
+
+All mix nodes in a deployment SHOULD use the same strategy type and parameters
+to ensure uniform emission patterns across the anonymity set.
+The configured path length `L` for cover packets
+MUST match the path length used for locally originated messages
+as defined in [Mix Protocol §6](./mix.md#6-pluggable-components).
+
+### 5.1 Cover Packet Transmission
+
+**Trigger:** The configured strategy schedules a cover emission.
+
+**[During Sphinx packet construction](./mix.md#85-packet-construction):**
+The mechanism constructs a cover Sphinx packet
+following the same construction procedure as a locally originated message,
+with the following differences:
+
+- The mix path is a loop path — the final hop routes the packet back to the originating node.
+- The origin protocol codec MUST be set to the cover traffic codec defined in [§5](#5-integration-with-the-mix-protocol).
+  This codec is recognized by the Mix Protocol during exit processing
+  to identify returning cover packets (see [§5.4](#54-cover-packet-reception)).
+- The application message content SHOULD be filled with cryptographically random bytes.
+  Random payloads provide defense-in-depth against partial path compromise
+  and ensure that cover packets remain indistinguishable from non-cover traffic
+  if the design evolves to support non-loop cover paths in the future.
+- If pre-computation is enabled, the pre-built cover packet is used directly without re-construction.
+
+**Wire format:**
+Cover packets use the exact Sphinx packet format defined in [Mix Protocol §8](./mix.md#8-sphinx-packet-format).
+No additional fields or framing are introduced.
+A cover packet on the wire is indistinguishable from a non-cover traffic packet,
+ensuring that intermediary nodes and external observers cannot classify packets as cover or non-cover.
+
+The cover packet is then transmitted to the first hop following the standard Mix Protocol transmission procedure.
+
+### 5.2 Non-Cover Slot Claim
+
+**Procedure:** `ClaimSlot() -> success`
+
+**Trigger:** The Mix Protocol needs to send a message or forward a packet and requires a slot from the budget.
+
+The mechanism atomically claims a slot from the pool using the following procedure:
+
+1. If `slots_remaining == 0`, the claim fails.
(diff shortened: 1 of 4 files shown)


docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,528 @@
...
+The caller then generates a DoS protection proof
+via `GenerateProof(binding_data)` ([Mix DoS Protection §8.2.1](./mix-dos-protection.md#821-proof-generation)),
+where `binding_data` is the packet-specific data as defined by the DoS protection mechanism.
+If no slot is available, the packet SHOULD be handled as follows to avoid hitting DoS protection limits:
+
+- **Locally originated messages**: queued for the next epoch.
+- **Forwarded packets**: dropped.
+
+When a slot is claimed for non-cover use,
+the pre-built cover packet associated with that slot MUST be discarded immediately.
```
jm-clius: Afaik the current zerokit library will not even allow you to generate this second proof in the same slot, as it violates RLN? Wondering if it might not be better to leave the idea of pregenerated proofs as simply a suggestion elsewhere in the spec and assume on-the-fly generation elsewhere for the sake of simplicity?

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,528 @@
...
+
+**Trigger:** The DoS protection mechanism signals the start of a new epoch
+via `OnEpochChange` ([Mix DoS Protection §8.2.3](./mix-dos-protection.md#823-epoch-change-notification)).
+The Mix Protocol MUST call `ResetEpoch` before processing any packets in the new epoch.
+
+The mechanism refreshes the slot pool for the new epoch:
+all remaining slots from the previous epoch are discarded,
+and a new pool of `R` slots is initialized.
+If pre-computation is enabled, the pre-built cover packets prepared during the previous epoch
+are loaded into the new pool.
```
jm-clius, line 256: For an RLN mechanism, depending on block time, valid RLN root window, etc. wouldn't precomputation run the risk of the root being invalid buy the time its intended epoch comes along? It may be worth stating the risk considerations of precomputation (proofs could be invalid), or perhaps leaving precomputation as a brief option top-level in the spec, but underspecified elsewhere (as suggested above). WDYT?

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,528 @@
...
+  cover emission is suppressed for the remainder of the epoch.
+  The resulting truncation makes inter-emission times near epoch end detectably different
+  from a true Poisson process.
+- **Epoch periodicity**: the budget resets every `P` seconds,
+  introducing periodic structure that does not exist in the continuous [Loopix](https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/piotrowska) model.
+
+Since forwarding traffic already introduces significant irregularity into total output,
+the practical difference between Constant-Rate and Poisson-Rate is minimal.
+Constant-Rate is preferred for its simplicity and exact pre-computation requirements.
+Poisson-Rate MAY be used when formal alignment with the [Loopix](https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/piotrowska) or [Nym](https://nymtech.net/nym-whitepaper.pdf) framework is required by a deployment.
```
jm-clius: Wondering if this is a strong enough reason for this to be included as an option in the spec? Perhaps better moved to the end as an appendix or future consideration?

jm-clius (approved):
This looks good to me!
Below I made some comments suggesting we underspecify the idea of proof precomputation, mostly to sidestep the complexity that introduces for RLN-based protection mechanisms (not only can a software bug lead to you slashing and deanonymising yourself, but because of root changes you might precompute invalid proofs). However, after reading through everything I agree that we probably want to keep proof precomputation as recommended default. Perhaps we can just RECOMMEND that all proofs be validated (again) before using them?

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: I was moving between constant and poisson rate initially and this got left here itself. 
But we can move to appendix, that makes sense. 
We can reanalyze once we finalize sender-generated proofs approach.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Since pre-computation is always for 1 epoch and we support the concept of max_epoch_gap to allow for other delays, this should not be a problem. 
Since we have mentioned pre-computation as optional to give flexibility. we can always turn-off/on based on actual behaviour observed in real network tests.

jm-clius: Wait, but that would account for the window of valid epochs, indeed. What about the window of valid _roots_ (against which merkle proof is created), which doesn't align with epoch?

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Was not aware of this. Thanks!
i assumed generate_proof can take the same message-Id which was used in generating cover packet proof which is being discarded. Then it should not be a problem. 
Will check and see.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: ah, right..that is another param to be considered. which means we need to check validity of the root just as a packet is sent out.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: looks like `generate_proof` has this option, which means we can reuse a `message_id`. 
Tested this with implementation and results of the test [here](https://github.com/logos-messaging/logos-delivery/pull/3807#issuecomment-4225312188)

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+
+The cover emission loop runs continuously as a background process.
+Emission timing is governed by the configured strategy ([`CoverTrafficConfig`](#55-data-structures)).
+
+**Algorithm: Cover Emission**
+
+> The following steps repeat continuously throughout each epoch:
+>
+> 1. Wait for the next emission event as determined by the configured strategy.
+> 2. If the strategy schedules an emission **and** the pool has available slots:
```
AkshayaMani: A few points could be made more explicit here:

`pool has available slots` — it would help to link this to the token bucket definition in §6, which clarifies that the pool is a bucket of `R` slots shared across all outgoing packets. A reader arriving at this loop without that context may not immediately understand what _"available"_ refers to.

`Send the pre-built packet to the first hop` — it would be worth explicitly stating that only the `packet` field of `PrebuiltCoverPacket` is transmitted on the wire, consistent with §5.1. The struct also contains `path`, `slot_id`, and `created_at`, which are internal state. Without this clarification, the boundary between the wire format and the internal data structure is left implicit.

Slot fire timing — §6.4 requires forwarded packets to call ClaimSlot() before the mixing delay, meaning a slot can be claimed but the packet has not yet been sent. Cover packets, by contrast, claim and send immediately in steps 2a–2b. The loop does not address how cover emission should behave when slots are claimed by in-flight forwarded packets that are still within their mixing delay window — _i.e.,_ whether those slots count as _"consumed"_ for the purpose of this check.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+the Mix Protocol instance MUST first call `ClaimSlot()` ([§5.2](#52-non-cover-slot-claim)).
+If no slot can be claimed, the message is queued for the next epoch.
+Otherwise, the Mix Protocol instance proceeds with
+[Sphinx packet construction](./mix.md#85-packet-construction).
+
+### 6.4 Packet Forwarding
+
+**[During Sphinx packet handling](./mix.md#86-sphinx-packet-handling):**
+When the Mix Protocol instance acts as an intermediary and receives a Sphinx packet to forward,
+it MUST first call `ClaimSlot()` ([§5.2](#52-non-cover-slot-claim)) before applying the mixing delay.
```
AkshayaMani, line 393: A couple of points around forwarded packet send timing and ordering could be made more explicit here:

Send timing — after `ClaimSlot()` succeeds, it would help to explicitly state that forwarded packets are sent independently when their mixing delay elapses, not gated on the cover emission schedule. Without this, the relationship between the mixing delay and the emission strategy timer is left implicit.

Slot Consumption" — since `ClaimSlot()` and wire transmission are decoupled by the mixing delay, it would help to clarify at what point a slot is considered consumed: at claim time or at transmission time.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+This ensures no two forwarded packets consume the same slot regardless of how their mixing delays overlap.
+If no slot can be claimed, the packet is dropped.
```
AkshayaMani, line 395: The cover loop fires continuously on a fixed schedule regardless of forwarded traffic load, which means cover implicitly takes priority: slots consumed by early cover emissions are unavailable to forwarded packets arriving later in the epoch, even if total traffic is within budget. Honest forwarded traffic from other nodes can silently disappear mid-path as a result.

For example, if forwarded traffic arrives late in the epoch: with `R = 100`, if the cover loop sends `10` packets in the first half of the epoch, only `90` slots remain for the second half — even if total real traffic is `95` (within budget). Five packets are dropped not because `R` is exceeded but because cover consumed their slots first.

Real network traffic is inherently variable — forwarded packets do not arrive uniformly throughout an epoch. The self-balancing model in §4 assumes an equilibrium that may not hold in practice, and it is worth evaluating through simulation whether the current design handles real-time traffic conditions reliably.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+making total output inherently irregular even though the cover emission rate is constant.
+
+At the recommended default rate, approximately `R / (1 + L)` cover packets are emitted per epoch.
+The originated cover emission rate is perfectly constant,
+so an adversary cannot distinguish epochs with heavy locally originated traffic from idle epochs
+by observing cover timing alone.
+
+**Tradeoff — periodic emission pattern:**
+Under low forwarding load, the fixed-interval emission pattern may be detectable
+via statistical analysis (_e.g.,_ autocorrelation).
```
AkshayaMani: This understates the problem.

Cover packets go out at fixed grid intervals. Forwarded packets go out at `t_arrive` + `exponential_mixing_delay` — always off-grid. Over enough observations, the periodic cover structure is recoverable through timing analysis regardless of forwarding load. At higher forwarding load, more cover packets are displaced and the grid becomes less regular, but the separation between grid-timed cover emissions and randomly-timed forwarded packets remains. 

The strategy therefore provides cover volume but not cover timing: an observer can identify which packets are cover and which are real traffic through timing analysis alone.

§11.3 — where all traffic types share the same timing grid — is the only mechanism that closes this. The current strategy does not achieve timing unobservability without it.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+  cover emission is suppressed for the remainder of the epoch.
+  The resulting truncation makes inter-emission times near epoch end detectably different
+  from a true Poisson process.
+- **Epoch periodicity**: the budget resets every `P` seconds,
+  introducing periodic structure that does not exist in the continuous [Loopix](https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/piotrowska) model.
+
+Since forwarding traffic already introduces significant irregularity into total output,
+the practical difference between Constant-Rate and Poisson-Rate is minimal.
+Constant-Rate is preferred for its simplicity and exact pre-computation requirements.
+Poisson-Rate MAY be used when formal alignment with the [Loopix](https://www.usenix.org/conference/usenixsecurity17/technical-sessions/presentation/piotrowska) or [Nym](https://nymtech.net/nym-whitepaper.pdf) framework is required by a deployment.
```
AkshayaMani: Given the timing separability concern in §7.1, Poisson-Rate's random emission times may be worth more serious consideration as an alternative. However, setting `λ_cover` to `R / ((1 + L) * P)` targets `100%` slot utilisation — Poisson variance means the budget is exhausted in roughly half of all epochs, triggering the exact cover suppression described in that section. A safety margin on `λ_cover` (_e.g.,_ `0.7 × R / ((1 + L) × P)`) would significantly reduce this, and is worth exploring.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+
+## 8. Initiating-Only Node Considerations
+
+Initiating-only nodes are short-lived with dynamic identifiers and do not forward traffic.
+They SHOULD NOT generate cover traffic,
+as cover traffic is only meaningful for nodes that participate continuously in the network with a stable identity —
+a briefly connected node has no sustained emission pattern to protect or contribute.
```
AkshayaMani, line 493: This explains why cover traffic is not practical for initiating-only nodes, but it would help to clarify what privacy properties are/aren't provided in its absence. Path anonymity and identity unlinkability remain, but an adversary observing the link to the first hop — or a malicious first hop itself — can directly observe session volume and timing. Making this explicit would give a clearer picture of the privacy model for these nodes.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+(see [§11.1](#111-cover-emission-rate-estimation) for deployment-specific sizing guidance).
+Non-cover traffic claims slots via `ClaimSlot()` ([§5.2](#52-non-cover-slot-claim)) as it arrives,
+making total output inherently irregular even though the cover emission rate is constant.
+
+At the recommended default rate, approximately `R / (1 + L)` cover packets are emitted per epoch.
+The originated cover emission rate is perfectly constant,
+so an adversary cannot distinguish epochs with heavy locally originated traffic from idle epochs
+by observing cover timing alone.
+
+**Tradeoff — periodic emission pattern:**
```
AkshayaMani: When pre-computation falls behind the emission schedule, §6.1 suggests falling back to on-demand generation. For the Constant-Rate strategy, this introduces a detectable timing irregularity relative to the expected grid interval, which has privacy implications. This might be worth acknowledging.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+
+Rather than pre-computing all cover packets in the previous epoch,
+implementations MAY batch pre-computation across epochs:
+an initial batch during epoch `N-1` to ensure cover packets are available at the start of epoch `N`,
+with subsequent batches computed incrementally during epoch `N` itself, staying ahead of the emission schedule.
+This reduces peak computational load and memory usage.
+
+### 9.2 Pool Status Tracking
+
+Implementations SHOULD maintain runtime counters for available slots, cover emissions, and non-cover consumptions.
```
AkshayaMani, line 533: The non-cover consumption counter records the exact number of real forwarded packets per epoch — precisely what traffic analysis attempts to measure. If this counter is exposed via external interfaces such as metrics endpoint, log file, or monitoring system, it becomes a traffic analysis interface accessible to anyone with read access. It would be worth specifying that these counters should be kept in-memory only and not exported via any external interface.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,532 @@
...
+Otherwise, the Mix Protocol instance proceeds with
+[intermediary processing](./mix.md#863-intermediary-processing).
+
+## 7. Recommended Strategies
+
+This section defines two cover emission strategies: Constant-Rate and Poisson-Rate.
+Both operate over the same `R`-slot pool and produce irregular total output
+because forwarding traffic claims slots at unpredictable times.
+They differ only in how cover emission is scheduled.
+Both strategies will naturally emit cover at approximately `R / (1 + L)` packets per epoch
```
AkshayaMani: In §4, `C` is defined as the total origination rate — cover and locally originated messages combined. Here,   `C` is implicitly treated as purely cover. This holds only if locally originated real messages are negligible. It would help to state this assumption explicitly, or clarify how the cover emission rate should be adjusted when the node is also actively sending real messages.

AkshayaMani (commented):
Great spec! Left a few comments with questions and suggestions.

docs/ift-ts/raw/mix-cover-traffic.md
AkshayaMani: Had a similar limitation with sender-generated proofs. RLN `acceptable_root_window_size` would help. It must be large enough to cover the number of membership updates expected during an epoch.

chaitanyaprem: Yea, i like the idea of just validating a proof before using it which seems to be possible in implementation.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Agreed, this is something that cannot be solved by any cover traffic model with rate limits. 
The only thing we can do is add improvements if we notice too much real traffic is getting dropped, one such improvement is schedule the forwarding traffic for subsequent epoch in case slots get exhausted. 

Also this is something that cant be noticed by simulations as well, because as you pointed out real traffic is very different than simulations. We can take this up as an enhancement once we notice how the network is behaving after deploying to testnet and gets used.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Initially i wanted to do something similar to what blend does and i kind of like that idea. i.e even for forwarding traffic or traffic originated locally from app layer always align them with scheduled time(similar to cover traffic). But i did not want to complicate it from the begining and kept it as [future work](https://github.com/logos-co/logos-lips/blob/81e6e87478915e37419e4fcc189e8a7fcff69744/docs/ift-ts/raw/mix-cover-traffic.md#113-pre-scheduled-emission-timing). 

If we think this an attack that would need fixing immediately, then we will have to make that future work i.e blend style cover schedule default.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Included this in the latest commit.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: This would require further analysis and would have impact on delays to be applied for each forwarding packet as those delays would have to be aligned with pre-scheduled timing. Hence maybe better to defer it. 
I will highlight this more and add points you have mentioned.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Did some analysis and also as per our discussion realized that we will have to generate more cover packets for poisson rate which adds to the load of RLN proof generation. Hence the takeaway from this is it is good to have a configurable fraction of cover rate `f` with default as `0.7` defined and implemented so that it can be tuned as we deploy and test cover traffic.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Moved poisson-rate to future considerations (§11.5) in latest commit.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Added §10.5 in latest commit. `cover_rate_fraction` (default 0.7) reserves ~30% headroom against forwarding spikes which reduces starvation risk. Further mitigations deferred to simulation.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Added `cover_rate_fraction` as normative config param (default 0.7) in §7, applies to both strategies. Poisson moved to §11.5 as future consideration.

chaitanyaprem:
@AkshayaMani also added an overview section in the spec now.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,694 @@
...
+The caller then generates a DoS protection proof
+via `GenerateProof(binding_data)` ([Mix DoS Protection §8.2.1](./mix-dos-protection.md#821-proof-generation)),
+where `binding_data` is the packet-specific data as defined by the DoS protection mechanism.
+If no slot is available, the packet SHOULD be handled as follows to avoid hitting DoS protection limits:
+
+- **Locally originated messages**: queued for the next epoch.
+- **Forwarded packets**: dropped.
+
+When a slot is claimed for non-cover use,
+the pre-built cover packet associated with that slot MUST be discarded immediately.
```
Cofson: The spec requires that the pre-built cover packet 'associated with that slot' MUST be discarded when a non-cover claim happens, but the `SlotPool` data structure (lines 291-294) tracks only `epoch` and `cover_queue`, with no explicit slot-to-packet mapping. §6.2 step 2a dequeues the head of `cover_queue` on cover emission, but `ClaimSlot()` for non-cover (§5.2) doesn't say how it interacts with `cover_queue` at all. Two implementations could diverge: one pops the head of `cover_queue` on every `ClaimSlot()`, another tracks a separate slot index and only discards when that slot is selected for cover. Worth either adding a `slot_index` field to `SlotPool` and tying it to claim order, or stating explicitly that `ClaimSlot()` for non-cover MUST also dequeue (and discard) the head of `cover_queue` if non-empty.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,694 @@
...
+
+### 6.1 At Epoch Boundary
+
+When the DoS protection mechanism signals the start of a new epoch,
+the Mix Protocol instance MUST invoke `ResetEpoch` ([§5.3](#53-epoch-boundary)) on the cover traffic mechanism
+to discard previous epoch state and initialize a new slot pool.
+
+**If pre-computation is enabled (RECOMMENDED):**
+The cover traffic mechanism pre-builds cover packets during epoch `N-1` for use in epoch `N`.
+For each slot to be pre-computed (at most `R`; see [§11.1](#111-cover-emission-rate-estimation) for sizing guidance),
```
Cofson: Cross-reference looks off: this points to §11.1 for sizing guidance, but §11.1 is 'Adaptive Cover Rate Fraction' (about tuning `f`), not pre-computation sizing. The actual pre-computation guidance lives in §9.1. Either fix the link to §9.1, or split out a dedicated §11.x for pre-computation sizing if a future-work item is intended.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,694 @@
...
+(individual packets remain classifiable by timing).
+
+Full timing unobservability requires the pre-scheduled emission timing enhancement
+([§11.3](#113-pre-scheduled-emission-timing)),
+where all traffic types share the same timing grid.
+Constant-Rate is the only strategy compatible with this upgrade path,
+as it requires deterministic emission times known at epoch start.
+
+**Characteristics:**
+Constant-Rate emits exactly `N = f × R / (1 + L)` cover packets per epoch (deterministic).
```
Cofson: The 'exactly N (deterministic)' framing here doesn't quite line up with §6.2 step 3 ('If no slots remain, suppress cover emission for the remainder of the epoch'). Under heavy forwarding (the failure mode §10.5 motivates `cover_rate_fraction` for), cover emission gets suppressed, so N is an upper bound rather than a deterministic count. Suggest a small reword along the lines of: 'Constant-Rate emits up to N = f × R / (1 + L) cover packets per epoch; the count is exact when forwarding does not exhaust the pool, which is the design intent behind `cover_rate_fraction = 0.7`.' That keeps the security claim in §10.4 honest about the conditional under which it holds.

Cofson (commented):
Genuinely solid spec, the self-balancing pool model is well argued and the iteration with AkshayaMani has tightened a lot. Three small things worth firming up before draft:

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: makes sense. will update it.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: looks like a left-over piece during some section reorg, will fix the cross ref.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: As of now we dont need a slot-to-packet mapping, any packet that is part of an epoch can be claimed by any actual packet. 
This mapping would be required in future once we have pre-scheduled emission introduced. 
But understood the gap and added a counter in the slotPool to track `slots_remaining` and use that to decide.

Cofson (approved):

PR #317, opened 2026-04-23
[draft / early WiP] Introduce Logos module interface / runtime / transport specs

kaiserd:
This PR introduces the current draft of the Logos module-system specs as a three-part split:

  - spec-module-interface.md
  - spec-module-runtime.md
  - spec-module-transport.md

This is a WIP draft intended to get early architectural and protocol feedback
from CCs I directly ask.
(Do not waste your time reading this yet.)

The current state is uneven. LOGOS-MODULE-RUNTIME and LOGOS-MODULE-TRANSPORT
are especially raw.

### major updates
* https://github.com/logos-co/logos-lips/pull/317#issuecomment-4643585267


docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+The implementation dCBOR-encodes `{cid}`, calls
+`call(call_user_data, "storage_module", {"method":"exists","params":{cid}}, len,
+&resp, &resp_len)`,
+decodes the response map, and writes `out_exists`. The signature mirrors
+the per-method function (section 2.4) but takes `logos_call_module_fn`
+instead of `logos_module_handle_t*`.
+`call_user_data` follows the same process-local callback context rules as
+the runtime-provided call-module hook.
+
+### 7.6 Generated Qt/QML Bindings
```
arnetheduck: everything qt should go to its own spec -> imagine writing a text ui instead of qt -> it will not need any of this stuff

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+freeze the real Storage module interface.
+
+### 1.2 Schema Metadata
+
+Every `.cddl` file MUST begin with a metadata block:
+
+```cddl
+; -- metadata --
+_module = "storage_module"
+_version = [1, 0]            ; [major, minor]
```
arnetheduck: the schema versioning infrastructure needs a hash -> a hashing scheme for the cddl needs to be defined -> possibly, we want a tree hash over the schema so that the consitutent parts of a message implicitly get a version - consider a date schema for example - used in other schema, the date "object" shape would be hashable / recognisable this way

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+**Reserved lifecycle names.** The lifecycle/runtime exports
+`logos_<module>_name`, `_schema`, `_version`, `_init`, `_destroy`, and
+`_dispatch` occupy the plain `logos_<module>_*` namespace. Per-method C
+functions therefore use the canonical form `logos_<module>_call_<method>` so
+schema methods such as `init`, `version`, or `destroy` do not collide with the
+required lifecycle symbols. This affects only the generated C symbol names;
+the wire method name remains the bare schema method name (for example
+`"version"`).
+
+**Type names.** CDDL named types map to `logos_<module>_<type_snake>_t`:
```
arnetheduck: a bit too much logos everywhere - I suspect we'll be better off giving more freedom to the "inital" part and instead reserving `logos_` for "system-defined" stuff / runtime - otherwise, there needs to be a cutout of reserved names in some other form

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+```
+
+Pattern: `LOGOS_<MODULE>_EVENT_<NAME_UPPER>` where hyphens become underscores.
+
+### 2.2 Primitive Type Mapping
+
+| CDDL type  | CBOR major type           | C type                          | Notes                              |
+|------------|---------------------------|---------------------------------|------------------------------------|
+| `bool`     | 7 (simple true/false)     | `bool`                          | `<stdbool.h>`                      |
+| `uint`     | 0 (unsigned integer)      | `uint64_t`                      | Always 64-bit for ABI stability    |
```
arnetheduck: an alternative is to simply forbid `uint` here and always be explicit in schemas

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,980 @@
...
+CDDL, for example `storage.started-event` or
+`storage.upload-progress-event`.
+
+The runtime delivers the event to all subscribers (local or remote). In
+socket mode, the module host translates `logos_publish_fn` calls into Event
+messages (tag 105) on all connections with matching subscriptions.
+If multiple caller connections have matching subscriptions, each connection
+MUST receive an Event message for its own matching subscription IDs.
+
+This publish path is for **narrow asynchronous notifications** only. Modules
```
arnetheduck: narrow?

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+MAY simplify the ABI by removing `_version()` and treating the schema as the
+sole source of version information.
+
+The benefits of keeping `_version()` in v0.1 are pragmatic:
+
+- **Cheap probing:** the runtime can query a small, stable symbol without
+  parsing the full schema text.
+- **Compatibility checks:** version lookup is easy during discovery,
+  loading, and connection setup.
+- **Operational clarity:** logs, diagnostics, and crash reports can report a
```
arnetheduck: the schema itself has a schema -> version would be part of that -> the question is how to update the schema of the schema

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+
+**Example — `storage.exists` response result:**
+```
+{"exists": true}
+```
+
+For methods with empty responses, the result is an empty map `{}`.
+For events, the `data` field encodes the event schema map.
+
+### 4.5 dCBOR Requirement
```
arnetheduck: todo: investigate relationship between dCBOR and [CDE](https://datatracker.ietf.org/doc/draft-ietf-cbor-cde/13/)

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1389 @@
...
+Method errors are carried in Response messages with the `error` field. A
+module returning `LOGOS_ERR_METHOD_NOT_FOUND` produces a Response error, not
+a protocol error.
+
+---
+
+## 5. Common Types (`logos_common.cddl`)
+
+The following types are shared by all modules and the transport protocol.
+They are defined in `logos_common.cddl` and available to all module schemas.
```
arnetheduck: this should be explicit - ie common may _exist_ as a module but if it's used by another module it should explicitly be referenced -> schema references need to be worked out

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,980 @@
...
+runtime provides its own event loop based on `poll()`/`epoll()`/`kqueue()`.
+
+### 7.2 Module Event Loop
+
+Module host processes run their own event loop:
+
+```
+while running:
+    msg = read_framed_cbor(socket)
+    response = logos_<module>_dispatch(msg)
```
arnetheduck: conveying asynchronicity here would probably be a good idea

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,980 @@
...
+The only runtime-level requirement is that any such bridge preserve the
+canonical module contract defined by LOGOS-MODULE-INTERFACE and the execution
+semantics defined by this runtime specification.
+
+---
+
+## 12. Streaming and Large Data (Future)
+
+This version of the spec does NOT address streaming or chunked transfer of
+large payloads. The default message size limit (16 MB, configurable) is
```
arnetheduck: why limit this?

docs/logos-core/raw/spec-module-transport.md
```diff
@@ -0,0 +1,628 @@
...
+The transport layer treats schema payload fields (`params`, `result`, and
+`data`) as opaque CBOR maps after envelope validation.
+Payload validation against module CDDL schemas is owned by the module dispatch
+layer defined in LOGOS-MODULE-INTERFACE.
+
+---
+
+## 3. Connection Lifecycle
+
+### 3.1 Connection Establishment
```
arnetheduck: consider that http has no "negotiation" of this sort - ie the way to address bidirectionality is really to defined the protocol in terms of its unidirectional behavior and combine - this makes it easier to define the protocol over a http transport that does not have a long-running session

docs/logos-core/raw/spec-module-transport.md
```diff
@@ -0,0 +1,628 @@
...
+subscription-id   = uint
+capability-token  = bstr .size 16
+schema-version    = [uint, uint]           ; [major, minor]
+```
+
+### 1.2 Message Types
+
+| Tag | Type        | Direction        | Purpose                            |
+|-----|-------------|------------------|------------------------------------|
+| 100 | Hello       | Both             | Connection handshake               |
```
arnetheduck: what's the interaction between these tag numbers and https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml ?

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,980 @@
...
+- Windows: `LoadLibrary(path)`
+
+After loading, the runtime looks up the required symbols:
+
+```c
+void* lib = dlopen("storage_module.so", RTLD_NOW | RTLD_LOCAL);
+
+/* Look up the name function to identify the module */
+typedef const char* (*name_fn)(void);
+name_fn get_name = (name_fn)dlsym(lib, "logos_storage_module_name");
```
arnetheduck: what's this for, really, since we already know the name here?

docs/logos-core/raw/spec-module-interface.md
kaiserd: done. Also removed comparisons from the runtime spec. (I can write a comparison doc later) 6c9aa151d0b68830161dd30f2270e6ccd9485ec3

docs/logos-core/raw/spec-module-runtime.md
kaiserd: I made this explicit. Also in the interface spec 6c9aa151d0b68830161dd30f2270e6ccd9485ec3

docs/logos-core/raw/spec-module-runtime.md
kaiserd: done  6c9aa151d0b68830161dd30f2270e6ccd9485ec3

docs/logos-core/raw/spec-module-transport.md
kaiserd: Yes :sweat_smile: using tags here was not a good idea (was concise but hacky). Changed to message types. 6c9aa151d0b68830161dd30f2270e6ccd9485ec3

docs/logos-core/raw/spec-module-runtime.md
kaiserd: Yes :pray:, good catch. The example was mixing two paths. If the runtime already knows the module name, it should look up the prefixed symbols directly. logos_module_name() is only for the unknown-library bootstrap path. 6c9aa151d0b68830161dd30f2270e6ccd9485ec3

docs/logos-core/raw/spec-module-interface.md
kaiserd: I'll introduce a Logos prelude that defines (among other things I'll add later):

  uint8 / uint16 / uint32 / uint64
  int8 / int16 / int32 / int64

  Bare uint, int, and nint will no longer allowed in schemas except inside the prelude definitions. This keeps
  ABI/codegen width explicit while avoiding repeated uint .size N / signed range expressions throughout schemas

docs/logos-core/raw/spec-module-interface.md
kaiserd: I split the naming rules by namespace: flat _module names stay for this revision, logos_ module names are reserved for Logos runtime/system modules, exact Logos-assigned flat module names are reserved, CDDL logos., logos-, and logos_ names are reserved for Logos-defined surfaces, and exported logos_* symbols are ABI-owned. 

What do you think about namespaced  module IDs like logos.storage in the CDDL schema?

docs/logos-core/raw/spec-module-interface.md
kaiserd: addressed in ef3e4b2fd4bd4d6026e305705454871780288233

docs/logos-core/raw/spec-module-interface.md
kaiserd: yes. I added a note. As I see it, dCBOR could become an instantiation of CDE.

docs/logos-core/raw/spec-module-interface.md
kaiserd: I will address this with the merkleization step as they are tightly coupled.

docs/logos-core/raw/spec-module-runtime.md
kaiserd: Addressed in ef3e4b2fd4bd4d6026e305705454871780288233

I moved this out of the runtime spec.
In the transport spec I put a limit, since otherwise you could mount resource exhaustion attacks.

docs/logos-core/raw/spec-module-transport.md
kaiserd: Addressed in ef3e4b2fd4bd4d6026e305705454871780288233 by separating directional message semantics from the stream binding. The current Unix/TCP stream binding and wire format remain unchanged, but the spec no longer treats the long-lived socket session as the protocol itself. HTTP-style/request-oriented bindings are noted as future separate interoperability specs.

Is this OK for now? Could spec an HTTP version later.

docs/logos-core/raw/spec-module-interface.md
kaiserd: I will address this with the merkleization step as they are tightly coupled.

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1461 @@
...
+
+This complete `storage` schema is illustrative only. It demonstrates the
+module-interface format; it is not the normative specification of the real
+Storage module API.
+
+```cddl
+; storage_module.cddl
+
+; -- metadata --
+_module = "storage_module"
```
arnetheduck, line 448: `-` vs `_` is an unnecessary difference here .. either "anything goes" or "pure C identifiers" - the in-betweens are messy

docs/logos-core/raw/spec-module-interface.md
arnetheduck: the important thing is to have a safe carve-out for the runtime that others shoudln't use, so that the core can add conflict-free names - it's also an open question whether logos.storage is a "system module" - ie its interface could be, if it's reimplementable by other systems but if it's very storage-specific, it's probably better that it's _not_ a system module (exposing core functionality might be a good candidate, ie an api for plugin management, but things that could be modules less so)

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1461 @@
...
+
+| CDDL type           | C type                      |
+|----------------------|-----------------------------|
+| `logos-result`       | `logos_result_t`            |
+| `logos-error-code`   | `logos_error_code_t`        |
+
+**Event constants.** Event names map to C `#define` constants:
+
+```c
+#define LOGOS_STORAGE_EVENT_UPLOAD_PROGRESS  "storage.upload-progress-event"
```
arnetheduck: what's the rule for producing these? why did event move around, and how do you prevent conflicts in the `-` -> `_` replacement?

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1461 @@
...
+| `bool`     | 7 (simple true/false)     | `bool`                          | `<stdbool.h>`                      |
+| `uint8`    | 0 (unsigned integer)      | `uint8_t`                       | Range 0..255                       |
+| `uint16`   | 0 (unsigned integer)      | `uint16_t`                      | Range 0..65535                     |
+| `uint32`   | 0 (unsigned integer)      | `uint32_t`                      | Range 0..4294967295                |
+| `uint64`   | 0 (unsigned integer)      | `uint64_t`                      | Range 0..2^64-1                    |
+| `int8`     | 0 or 1 (signed integer)   | `int8_t`                        | Range -128..127                    |
+| `int16`    | 0 or 1 (signed integer)   | `int16_t`                       | Range -32768..32767                |
+| `int32`    | 0 or 1 (signed integer)   | `int32_t`                       | Range -2^31..2^31-1                |
+| `int64`    | 0 or 1 (signed integer)   | `int64_t`                       | Range -2^63..2^63-1                |
+| `float64`  | 7 (double-precision)      | `double`                        |                                    |
```
arnetheduck: floats are problematinc for determinism so we have to tread carefully here and consider whether they're desireable

docs/logos-core/raw/spec-module-interface.md
kaiserd: I'll simplify this. The idea was to allow grouping by EVENT in autocomplete. But the rule adds complexity.

docs/logos-core/raw/spec-module-interface.md
kaiserd: I'll update it to pure C and add a comment why this was chosen over more canonical CDDL (-> 1:1 mapping)

docs/logos-core/raw/spec-module-interface.md
kaiserd: I'll look into this again, while doing the merkle spec. I'll have an update soon. Most likely will remove floats to avoid complexity in the merkleization.

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,2430 @@
...
+
+A module that needs to call another module can do so through a runtime-
+provided callback, without linking the SDK or implementing the transport
+protocol's client side:
+
+```c
+/* Provided by the runtime to the module after _init() */
+typedef void (*logos_free_response_fn)(void* user_data, void* ptr);
+
+typedef logos_result_t (*logos_call_module_fn)(
```
arnetheduck: these really belong in `_init` - ie otherwise you get into trouble with sequencing, ie when can these be called - what happens if they're called mid-processing etc.

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,2430 @@
...
+dependency-graph queries, capability decisions, action prompts, or persistent
+configuration storage.
+Those belong to Package Manager, Capability Authority, Module Launcher,
+runtime-host, or deployment specifications.
+LOGOS-MODULE-SECURITY-CONSIDERATIONS records related threat analysis and
+hardening guidance.
+
+### 8.1 Module Initialization Inputs
+
+This revision defines the lifecycle initializer as `_init(void)`.
```
arnetheduck: it has crossed my mind that configuration should be part of the module schema and passed in here _or somehwere_ - most plugin systems end up having this need regardless and this plugins can be configured from a "smart" editor, assuming the config schema makes sense - it's not urgent for starters but it's likely we'll end up with a standardised config interface.

arnetheduck: this ties in with callbacks btw .. probably we'll end up with something like `struct {(void set_call_module*)(...); ..; const char *config_data; ...} ModuleConfig;` where the various pointers etc can be `NULL` for modules that don't care or that don't require the return interface ..

also, more likely than a generic "call" interface might be that modules request an "interface" to call specific modules, ie `callback get_interface_for(const char* module_name)`  or `callback get_interface_for(const char* capability)` and use these more specific callbacks instead of a generic callback routine .. anyway, this is just a loose thought that needs more development

kaiserd: > configuration should be part of the module schema and passed in here or somewhere

Yes! I fully agree, I was already experimenting with this. I'll have a new larger commit soon,
wrapping up a first version of security / access policies, a capability authority module, and a module launcher module.

docs/logos-core/raw/spec-module-transport.md
```diff
@@ -0,0 +1,815 @@
...
+
+schema-commitment = {
+    commitment_model:  semantic-commitment-model-revision,
+    schema_root:       schema-root,
+    hash_profile:      hash-profile-id,
+    hash_suite:        hash-suite-id,
+}
+
+hello = {
+    0:              message-kind-hello,
```
arnetheduck: Usually it's easier to put the tag outside of the payload as part of the framing ([together](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value) with length) - in your decoder, you want to know what type of message it is before decoding the fields so if you put it inside you either have to make two passes over the data _or_ write a custom decoder that picks out only the tag before decoding the rest - with CBOR, the order of fields is not guaranteed so the tag might end up at any byte offset in the stream - CDE fixes this by lexicographically ordering things so a 0 _should_ end up first but it's still messy - ie in CDE encoding, the first byte is the "map" marker so the second byte becomes the tag.

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+}
+
+peer_info = {
+    id: tstr,
+    addrs: [* tstr],
+    ? name: tstr,
+}
+```
+
+### 1.7 Type Restrictions
```
arnetheduck, line 382: this needs to point to the deterministic cddl / cbor spec

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+- `logos_<module>_name()` returns the module name as a static string that
+  remains valid for the library lifetime.
+- `logos_<module>_schema()` returns the canonical CDDL schema text as a
+  static string that remains valid for the library lifetime.
+- `logos_<module>_version()` returns a concise schema version string such as
+  `"1.0"`.
+- `logos_<module>_init()` is called once after loading.
+  It returns `0` on success or a nonzero Logos error code on failure.
+- `logos_<module>_destroy()` is called once before unloading.
+- `logos_<module>_free()` releases dynamic memory returned by this module
```
arnetheduck:
```suggestion
- `logos_<module>_free()` releases dynamic memory allocated by this module
```

docs/logos-core/raw/spec-module-transport.md
arnetheduck: coming back to the tag question, putting the tag inside the message is somewhat problematic since it no longer appears in a fixed location in the data stream meaning that in order to interpret the message, you must first decode it as a "generic" cbor message.

we can:
* put the tag in the frame header instead together with length - classic [TLV](https://en.wikipedia.org/wiki/Type%E2%80%93length%E2%80%93value), simple
* go back to cbor tags - this is the "native" cbor way to deal with this problem - the tag then has a fixed location in the byte stream (beginning) and is therefore easy to pick out _before_ decoding the rest of the message aka message validation and memory allocation can be done efficiently and invalid messages can be rejected earlier
* use an array `[tag, ...]` - similar but less extensible, ie not entirely appropriate for messages like hello that may get extended or change over time

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+that remains necessary after lifecycle `_init()` has succeeded.
+
+Successful lifecycle `_init()` therefore means:
+
+- the module has loaded correctly into the runtime
+- the runtime or module host may invoke schema-defined methods through the
+  native per-method C entry points
+
+It does **not** mean every schema-defined method must already succeed.
+Schema-defined methods MAY still return `LOGOS_ERR_NOT_READY` until the
```
arnetheduck: this is probably a bad idea, overloading the same error code for subtly different meanings - an api should have its own errors for its own behaviors while "runtime-level" errors should be distinct.. also, per the other comments, when `init` returns a context, this context is required to call any other function -> there is by construction no way to call the methods when they're not ready - the context however might be invalid which is a "different" kind of error.

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+storage.upload_done_event = {
+    cid: tstr,
+}
+
+storage.started_event = {}
+```
+
+---
+
+## 2. CDDL-to-C Canonical Mapping
```
arnetheduck, line 505: the threading model deserves to be repeated here, at least briefly .. in some ways, it makes more sense here than in the runtime doc .. in fact, in the runtime, the threading might actually be an implementation detail that is not necessarily part of the spec, or at least it's something a module writer should not have to read the runtime spec to understand (if you're writing a module, you don't want to know how the runtime works, you just need to understand the surface between the module and the runtime)

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+    size_t              detail_len;
+} logos_result_t;
+
+/* -- Call context -- */
+
+/*
+ * Opaque inbound-call context supplied by the runtime or module host.
+ * Module implementations may ignore this pointer.
+ */
+typedef struct logos_call_context logos_call_context_t;
```
arnetheduck: This looks wrong - the context is something that `_init` should return, ie it's a value that the runtime receives from `init` (that `init` returns) and that's given to the module by the runtime in every function call - it's meant to carry instance state that the module needs and constructs during init (and frees during destroy)

Notably, if we call _init twice we will generate two instances of the same module - unusual perhaps but valid .. once we get into the territory of providing configuration for the module, this becomes more relevant (because now we can init multiple instances of the same module but with different configs).

As mentioned elsewhere, the runtime context that we _give_ to the module should be part of the init call

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+    version_mismatch:  7,
+    not_ready:         8,
+    cancelled:         9,
+)
+
+; -- result type --
+logos_result = {
+    code: logos_error_code,
+    ? message: tstr,
+    ? detail: bstr,
```
arnetheduck: detail should an context-defined cbor structure of arbitrary type so that "rich" errors can be transmitted

docs/logos-core/raw/spec-module-transport.md
```diff
@@ -0,0 +1,815 @@
...
+
+### 5.3 Async Operations Pattern
+
+1. Caller subscribes to progress/completion events.
+2. Caller sends Request.
+3. Callee responds with ack.
+4. Callee publishes Event messages as the operation progresses.
+5. Caller unsubscribes when done.
+
+There is no "async method" concept. All methods are request/response.
```
arnetheduck, line 750: well there is - ie responses are out of order and processed concurrently, that's as async as it gets? the interesting question with respect to async here is if the first event is allowed to happen _before_ the response to the "subscribe" message.

with subscribe systems, there's always an interesting race condition: if your events are "diffs" (ie value changed to X), how do you synchronize getting the "current state"? if you get the state before subscribing, it's stale and if you get it after, you need an identifier in the event that shows which state it refers to - in this sense,

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+    cid: tstr,
+}
+```
+
+Events are NOT tied to specific methods. Any caller that subscribes to an
+event receives it whenever the module publishes it. Event subscription is
+managed by the runtime (see LOGOS-MODULE-RUNTIME section 4) and the transport
+protocol (see LOGOS-MODULE-TRANSPORT section 5).
+
+Events are an asynchronous one-way notification mechanism with schema-defined
```
arnetheduck: this doesn't really make sense for in-process modules / the C API - ie the runtime "publishes" events by calling a callback synchronously and there is no asynchronicity here - neither are there mechanisms for queuing or dropping events - if the module is slow to process them, the runtime slows down (unlike a distributed module where a consumer might be slow leading to queue saturation).

if we make this constraint here, it effectively means that we run events on a separate thread with an intermediate queue that disconnects the runtime from the "natural" backpressure that a callback offers in this regard - this is probably an an unnecessary complication.

docs/logos-core/raw/spec-module-interface.md
```diff
@@ -0,0 +1,1661 @@
...
+
+A method with no input uses an empty map: `storage.space_request = {}`.
+A method whose only output is success/failure uses an empty response map:
+`storage.destroy_response = {}`. Success is indicated by `logos_result_t.code
+== LOGOS_OK`; the empty response means no additional data.
+
+Map keys MUST be bare CDDL identifiers (not quoted strings). Key names are
+used directly as C parameter names and deterministic CBOR map keys.
+
+### 1.5 Event Declarations
```
arnetheduck, line 329: how do you subscribe to events in the C API?

docs/logos-core/raw/spec-module-runtime.md
```diff
@@ -0,0 +1,2430 @@
...
+| `module_instance_id` | Opaque identity for a live module instance. | `tstr .size (1..128)` | Runtime or provider scope that owns the live instance. |
+| `schema_namespace` | Structural schema namespace for a module contract. | `tstr .size (1..128)` | Schema model, not runtime addressing. |
+| `schema_commitment` | Commitment to the structural schema contract. | Commitment record. | Hash and commitment profile scope. |
+
+```cddl
+; -- metadata --
+_module = "logos_runtime_control"
+_version = [1, 0]
+
+logos.runtime_control.module_name = tstr .size (1..64)
```
arnetheduck: oof ugly to repeat this all the time .. wonder if something like https://cbor-wg.github.io/cddl-modules/draft-ietf-cbor-cddl-modules.html  can help

kaiserd:
Larger update.

- addressed the feedback in the comment above, except for hierarchical CDDL namespaces, because the relevant IETF
draft has expired and is no longer a valid normative basis;
- introduced interfaces as a distinct contract kind, including interface implementation declarations, exact interface contract views, provider selection through interfaces, supporting schemas, and conformance coverage;
-  unified the participant model around modules: applications, presentation components, consumers, providers, and privileged system services use ordinary module identity, lifecycle, routing, and authority semantics rather than separate participant classes;
- improved Transport session establishment, ticket redemption, authorization failures, event subscriptions,
cancellation, and endpoint protection;
- clarified schema construction, deterministic encoding, C mapping, commitments, route access, and size bounds;
- expanded exact-byte and state-transition conformance coverage and regenerated affected schema roots;
- aligned terminology, ownership, and security boundaries across the specification set.

This PR also introduces six new specifications (this are still in early stage):

- Capability Authority
- Configuration
- Module Loader
- Package Manager
- Security Considerations
- System BCP

With this, the intended feature set (so far) is covered. Next steps: PoC, condensing specs, testing, ...

PR #318, opened 2026-04-24
de-MLS clarifying edge cases

seugu:
The following PR includes edge cases of the current RFC

added: 
- keyPackatge buffering
- self removal and epoch steward self removal edge case clarifications
- Better handling right after the Emergency criteria proposal
- adding retry_value for having another verifiable set for non-election methods 
- relaxed RFC by moving actions to local

Thanks to @seemenkina for raising these issues, making the RFC solid.


docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -390,15 +401,86 @@ since allowing bias could enable a malicious participant to manipulate the list
...
+
+If the re-election proposal is finalized with YES, the new `steward list` is installed and Layer 1 is applied again.
+Otherwise, if the proposal is finalized with NO, `retry_round` MUST be incremented
+and the re-election process MAY be repeated until `max_reelection_attempts` is reached.
+Note that `max_reelection_attempts` is the parameter that is set during group creation.
+
+If no new `steward list` can be established after exhausting `max_reelection_attempts`,
+the system enters a steward deadlock condition, and Layer 3 MUST be activated.
+
+##### Layer 3 - Bounded retry failure
```
seemenkina:
```suggestion
##### Layer 3 — Anti-deadlock ECP
```
"Bounded retry failure" names the trigger (Layer 2 exhausted retries); the layer's actual content is the anti-deadlock ECP. The new name reads in parallel with the other layers ("Local steward rotation", "Re-election" → "Anti-deadlock ECP") and matches the vocabulary already used in the Consensus Types section.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -390,15 +401,86 @@ since allowing bias could enable a malicious participant to manipulate the list
...
+ensuring minimal intervention while maintaining system continuity.
+
+##### Layer 1 - Local steward rotation
+Layer 1 ensures that a finalized voting proposal is committed by locally rotating over the active `steward list` in deterministic order.
+
+A steward is eligible to act as the `epoch steward` if there is no malicious behavior as defined in [Steward Violation List](#steward-violation-list).
+Otherwise, the `epoch steward` MUST be considered ineligible and MUST be locally assigned a low score by members.
+Upon detecting an `epoch steward` with lower peer score than `threshold_peer_score`,
+members MAY initiate an Emergency Criteria Proposal (ECP) to remove the current `epoch steward`.
+Once removal is accepted, the `backup steward` becomes the new `epoch steward`.
```
seemenkina: I suggest to change tot something like this: 

A steward is eligible to act as the epoch steward if it is a current group member and not pending removal. Misbehavior per the Steward Violation List does not affect eligibility directly; it decrements peer
score per the Peer Scoring section, and a steward becomes ineligible only once a ECP finalizes the removal.

The reason: 

The original conflates two distinct properties — availability for rotation and misbehavior. A misbehaving-but-not-yet-removed steward is still eligible to commit; they're just on track for removal. The Layer 1 routing rule needs to depend on availability only; otherwise members would diverge whenever they observed a misbehavior at slightly different times. Splitting the two cleans the layered model and keeps scoring logic where it belongs.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -390,15 +401,86 @@ since allowing bias could enable a malicious participant to manipulate the list
...
+
+Until such removal is finalized, members MUST skip the ineligible steward and continue walking the `steward list`.
+The first eligible steward in the list SHOULD produce the commit.
+If the `epoch steward` is ineligible, the `backup steward`, or the next eligible steward in the list,
+MAY commit without requiring an Emergency Criteria Proposal.
```
seemenkina: Based on previous comment probably we need to adjust this sentence because the walk is conditioned on ineligibility, not on a pending ECP. 

Suggest something like this: 

When the nominal epoch steward is ineligible, members MUST walk the steward list in deterministic order and accept the commit produced by the first eligible steward. The backup steward or any subsequent
eligible steward MAY commit without an Emergency Criteria Proposal.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -390,15 +401,86 @@ since allowing bias could enable a malicious participant to manipulate the list
...
+
+Even in the presence of malicious or silent stewards,
+Layer 1 preserves liveness by progressing through the `steward list`
+until an eligible steward is found to produce the commit.
+This guarantees that isolated failures do not block the system.
+If no eligible steward can be found across the entire `steward list`,
+or if the issue cannot be resolved within Layer 1,
+the protocol escalates to Layer 2 for coordinated recovery.
```
seemenkina: Even when individual stewards are silent or have been removed, Layer 1 preserves liveness by walking the steward list until an eligible steward produces the commit. Misbehaving stewards continue to participate
in rotation until accumulated scoring triggers ECP, at which point removal makes them ineligible and Layer 1 walks past them on subsequent rounds. If no eligible steward exists across the entire list, the protocol escalates to Layer 2. 

---

Sorry for the a lot fixes, but my suggestion that Layer 1's behavior becomes a single rule ("skip ineligible, walk to next") with a clean definition of "ineligible".

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -489,6 +569,33 @@ Competing commits that contain the same deterministic proposal sequence
...
+
+A `steward` MUST NOT produce a commit that includes its own removal.
+This is not a protocol violation but an inherent constraint of [MLS RFC 9420](https://datatracker.ietf.org/doc/rfc9420/):
+a member cannot apply a commit that removes itself from the group;
+otherwise, the resulting group key would be accessible to the removed member again,
+which is a contradiction of removal.
+
+A member's request to leave the group is not subject to voting.
+Any such removal voting proposal MUST be automatically treated as finalized with a YES outcome
+and MUST be processed by the epoch steward in the subsequent commit.
```
seemenkina:
```suggestion
A member's request to leave the group is not subject to voting. 
A removal voting proposal is auto-finalized YES only if its sender, proposal_owner, and RemoveMember.target all reference the same identity;
members MUST reject any proposal claiming auto-YES status that fails this check and MAY apply a peer-score penalty against the sender. Qualifying proposals MUST be processed by the epoch steward in the
subsequent commit.
```

Closes the unilateral-removal vector that auto-YES otherwise opens — without the gate, any member can remove any other.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -489,6 +569,33 @@ Competing commits that contain the same deterministic proposal sequence
...
+- The `epoch steward` waits until it has committed all currently finalized voting proposals,
+or if none are pending, MAY initiate a self key rotate request to trigger an epoch transition and naturally offload its role.
+
+- The `epoch steward` creates its own remove proposal, which is automatically finalized with a YES outcome,
+but deliberately does not commit it, leaving it to the subsequent epoch steward to include in the next commit.
+This constitutes a valid exception to the requirement that finalized voting proposals MUST be committed;
+the epoch steward MUST NOT be penalized for this omission.
+Under this approach, removal requires only a single commit by the subsequent epoch steward.
+
+Implementations MAY choose either approach. Both are compliant with this specification.
```
seemenkina, line 587:
```suggestion
Implementations MAY choose either approach. Both are compliant with this specification.

If the self-removing steward is the sole eligible member of the active steward list, 
neither approach above suffices — no successor exists to commit the removal. 
The protocol MUST trigger Layer 2 re-election with the leaver excluded from the candidate pool; 
the newly elected epoch steward then includes the removal in its first commit.
```

Links Self-Removal to the layered recovery model for the no-successor case — without this paragraph, a reader has no answer for "what if the leaver is the only steward?"

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -489,6 +569,33 @@ Competing commits that contain the same deterministic proposal sequence
...
+Two  approaches are valid at the application level:
+
+- The `epoch steward` waits until it has committed all currently finalized voting proposals,
+or if none are pending, MAY initiate a self key rotate request to trigger an epoch transition and naturally offload its role.
+
+- The `epoch steward` creates its own remove proposal, which is automatically finalized with a YES outcome,
+but deliberately does not commit it, leaving it to the subsequent epoch steward to include in the next commit.
+This constitutes a valid exception to the requirement that finalized voting proposals MUST be committed;
+the epoch steward MUST NOT be penalized for this omission.
+Under this approach, removal requires only a single commit by the subsequent epoch steward.
```
seemenkina:
```suggestion
The offload mechanism is implementation-defined — 
examples include queuing the removal for the next eligible steward to commit on their next turn, 
or triggering an epoch transition (e.g. via a key rotation) so that role rotation hands off naturally. 
The originating steward MUST NOT be penalized for omitting its own removal from its commits — 
this is a recognized exception to the rule that finalized voting proposals MUST be committed.
```

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seemenkina, line 609: I suggest to add next sentence after: 

All three violation types are detected locally by members; each detection contributes a peer-score decrement per the Peer Scoring section. Removal occurs only after the steward's accumulated score drops below threshold_peer_score, triggering a ECP per the Consensus Types section.

-- 
As the list defines three violations but never says how they're handled. Without this paragraph, a user could reasonably assume each violation triggers its own ECP. Centralizing the handling rule (local detection → scoring → threshold ECP) avoids restating it under each item, and makes the connection to Layer 1's eligibility model explicit.

seemenkina:
```suggestion
the commit validation service MUST select the epoch steward's commit when present;
otherwise (e.g. during Layer 3 recovery mode, where any member MAY commit) 
the commit with the lexicographically smallest committer ID is selected, avoiding different state forks.
```

seemenkina: The original sentence  never names the case where no epoch-steward commit can exist — Layer 3 recovery mode.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -544,9 +650,10 @@ The exact scoring rules, recovery mechanisms, and escalation criteria are left f
 
 ## Timer-Based Anti-Deadlock Mechanism
```
seemenkina: I think this section rn doesn't make sense after we presented layers  because it confused the trigger with the response. The timer is the trigger for Layer 1; the deadlock case lives in Layer 3 and is reached via Layer 2 exhaustion, not via a separate deadlock-detection timer. The replacement keeps the protocol-relevant rule (when to start/reset, what expiry means) and defers the duration value + buffer policy to per-group configuration 

---

## Inactivity Timer

Each member MUST maintain local timers to detect when expected protocol events fail to occur within a bounded time. The protocol relies on inactivity detection in three contexts:

1. Commit inactivity: The epoch steward does not produce a commit referencing a finalized voting proposal. On expiry, the member treats the epoch steward as inactive and proceeds with Layer 1 rotation.
2. Recovery inactivity: During an active recovery window (Layer 2 re-election in progress, or Layer 3 recovery mode set), a separate, typically shorter inactivity duration SHOULD apply so retries do not burn a full epoch.
3. Voting inactivity: A submitted update request (e.g. an add or remove) does not progress to an open consensus session within the expected window. Members MAY initiate or re-submit the corresponding voting proposal directly.

Each timer's duration and any tolerance buffer for P2P timing variance are configured per group. Escalation beyond Layer 1 follows the Three-Layer Steward Protection Mechanism.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -314,6 +321,10 @@ reflecting the expectation of active participation from members in the decentral
 If the proposal returns YES, a score penalty MUST be applied to the targeted member or steward
 by decreasing their peer score, and a score reward MUST be granted to the creator of the proposal;
 if the proposal returns NO, a score penalty MUST be applied to the creator of the proposal.
+If the targeted member is the `epoch steward`,
+the `backup steward` MUST assume the `epoch steward` role and MUST include the removal of the former `epoch steward`
+in the subsequent commit without requiring an additional voting round,
+as the finalized emergency criteria proposal serves as sufficient authorization for the removal.
 `Proposal.payload` MUST include evidence of dishonesty as defined in the Steward Violation List,
 along with the identifier of the malicious member or steward.
 This proposal can be created by any member in any epoch.
```
seemenkina: 3. Emergency criteria proposal (ECP): A consensus action carrying a violation_type field that discriminates between
(a) member removal, where Proposal.payload MUST include the target identifier and supporting
evidence per the Steward Violation List;
(b) protocol deadlock, where no specific target exists and recovery is handled per Layer 3.  Any member MAY create an ECP in any epoch. On YES, members MUST enter the
freezing phase immediately, bypassing the inactivity timer, so the consequent commit lands without waiting a full cycle and a peer-score reward MUST be granted to the creator of the proposal; 
on NO, a peer-score penalty MUST be applied to the creator to deter abuse.

--- 
So such changes  cover next thoughts: 
- Adds the violation_type discriminator (member removal vs. deadlock) 
- Adds force-Freezing-on-YES — without this, removal/recovery waits a full inactivity cycle.
- The "backup steward MUST assume role and include removal" sentence is dropped because Layer 1's rotation walk already handles successor selection — restating it here duplicates Layer 1.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: Added [here](https://github.com/logos-co/logos-lips/pull/318/commits/b2cc6da3ef09efa0f4464ccd3e3510c3145c0d19)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: addressed [here](https://github.com/logos-co/logos-lips/pull/318/commits/a4ee682a580b270f2075d41634b80820bb240c53)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: addressed [here](https://github.com/logos-co/logos-lips/pull/318/commits/77d899c38c7a5451736ff540c90b13ec6a351b49)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: yes, addressed [here](https://github.com/logos-co/logos-lips/pull/318/commits/5709a2182860dba83712b2274a6c5be557bd1d9b)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: I had similar feelings this makes clear, [here](https://github.com/logos-co/logos-lips/pull/318/commits/54cc28c42415bfc392e2878a229eef1934cb7792)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: Changed as you point out [here ](https://github.com/logos-co/logos-lips/pull/318/changes/1e97bbd869288675c4cba166c463c8b36351d416)

Cofson (commented):
Mostly already covered by previous review. Spotted a few things that look like incomplete edits or small typos:

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -484,11 +554,36 @@ Therefore, proposal equivalence alone does not guarantee state equivalence.
 If multiple valid commits contain the identical deterministic proposal sequence,
 the commit validation service MUST select first, if there is `Epoch steward`,
 otherwise the commit whose `committer ID` is lexicographically smallest (according to canonical ordering)
-as the single valid output, thereby avoiding different state forks.
+the commit validation service MUST select the epoch steward's commit when present;
+otherwise (e.g. during Layer 3 recovery mode, where any member MAY commit) 
+the commit with the lexicographically smallest committer ID is selected, avoiding different state forks.
 Competing commits that contain the same deterministic proposal sequence
```
Cofson, line 561: Looks like an incomplete edit — when the new selection rule was applied here, the previous three lines (`MUST select first, if there is Epoch steward, otherwise the commit whose committer ID is lexicographically smallest (according to canonical ordering)`) were not removed. The result reads as two rules concatenated without a sentence break. The old prefix sentence should be dropped so only the new rule remains.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -308,15 +315,13 @@ The `Proposal.payload` field MUST represent the ordered identities of the propos
...
-3. `Emergency criteria proposal`: If there is a malicious member or steward,
-this event MUST be finalized through a governance vote,
-reflecting the expectation of active participation from members in the decentralized governance process.
-If the proposal returns YES, a score penalty MUST be applied to the targeted member or steward
-by decreasing their peer score, and a score reward MUST be granted to the creator of the proposal;
-if the proposal returns NO, a score penalty MUST be applied to the creator of the proposal.
-`Proposal.payload` MUST include evidence of dishonesty as defined in the Steward Violation List,
-along with the identifier of the malicious member or steward.
-This proposal can be created by any member in any epoch.
+3. `Emergency criteria proposal (ECP)`: A consensus action carrying a `violation_type` field that discriminates between
```
Cofson, line 318: Both `Emergency criteria proposal` (here, 321, 329, 341) and `Emergency Criteria Proposal` (424, 428, 432, 461, 466, 472, 611) appear throughout. Worth a single pass to normalise — Title Case probably reads better since it is effectively a defined term that gets the `(ECP)` short form.

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
```diff
@@ -279,7 +280,7 @@ Decentralization has already been achieved in the previous section.
 However, to improve availability and ensure censorship resistance,
 the single steward protocol is extended to a multi steward architecture.
 In this design, each epoch is coordinated by a designated steward,
-operating under the same protocol as the single steward model.
+operating under the similar protocol as the single steward model.
```
Cofson: A few small copy fixes spotted during reading, grouping them here rather than spreading them inline:

- line 283: "the similar protocol" → "a similar protocol" (or revert to "the same protocol", which read fine pre-PR)
- line 321: missing closing backtick — `` `Emergency criteria proposal (ECP)` `` in any epoch
- lines 424, 611: "a `Emergency Criteria Proposal`" — vowel sound takes "an"
- line 432: "triggers `Emergency Criteria Proposal (ECP)`" — missing article ("triggers an")

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: All proposals are title case now, [here](https://github.com/logos-co/logos-lips/pull/318/commits/d4c481593d9a2cf6db1ae9c7bbedb933e7652393)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: good catches thank you! addressed [here](https://github.com/logos-co/logos-lips/pull/318/commits/b4e9b69fa1c5f413dd6337500c755465b76b4670)

docs/ift-ts/raw/decentralized-mls-offchain-consensus.md
seugu: Oh I missed this, thank you here is addressing [commit](https://github.com/logos-co/logos-lips/pull/318/commits/efa4c8a21bc4ba766449fee23b418ffd2d3d50d0)

Cofson (approved):
All three follow-ups from my last pass are in (Title Case normalised, lexicographic-tiebreaker selection rule completed, copy fixes applied). LGTM from my side.

seemenkina (approved):

PR #322, opened 2026-04-29
Clarify lower bound enforcement and update logic for waiting time

tinniaru3005:
Refines the lower-bound section to clearly define bound/timestamp semantics


docs/ift-ts/raw/logos-service-discovery.md
```diff
@@ -1325,38 +1325,59 @@ repeatedly request new tickets hoping for better waiting times,
...
-cannot be smaller than a previous waiting time `w_1` at time `t_1`
+MUST NOT be smaller than a previous waiting time `w_1` at time `t_1`
 (where `t_1 < t_2`) by more than the elapsed time:
 
 ```text
 w_2 ≥ w_1 - (t_2 - t_1)
 ```
 
-Thus registrars MUST maintain lower bound state for:
+For each `service_id_hash` and each `IP`, the registrar SHOULD maintain:
```
jm-clius:
```suggestion
For each `service_id_hash` and each `IP` prefix in the IP tree, the registrar MUST maintain:
```

docs/ift-ts/raw/logos-service-discovery.md
```diff
@@ -1325,38 +1325,59 @@ repeatedly request new tickets hoping for better waiting times,
...
 
-- Each service in the cache: `bound(service_id_hash)` and `timestamp(service_id_hash)`
-- Each IP prefix in the IP tree: `bound(IP)` and `timestamp(IP)`
+- `bound`: the last issued waiting time (`w_1`) when bound value was updated
+- `timestamp`: the time at which `w_1` was issued (`t_1`)
```
jm-clius: I would make it explicit that we therefore maintain `bound(service_id_hash)` and `bound(ip)` so we can use these symbols exactly below. Same for `timestamp`.

docs/ift-ts/raw/logos-service-discovery.md
```diff
@@ -1325,38 +1325,59 @@ repeatedly request new tickets hoping for better waiting times,
...
+appearing in advertisements associated with non-expired `WAIT` tickets.
 
-**How SHOULD lower bound be calculated for service IDs:**
+**How SHOULD lower bound be calculated**
 
-When new `service_id_hash` enters the cache, `bound(service_id_hash)` is set to `0`,
-and a `timestamp(service_id_hash)` is set to the current time.
-When a new ticket request arrives for the same `service_id_hash`,
-the registrar calculates the service waiting time `w_s` and then applies the lower-bound rule:
+When a new ticket request arrives, the registrar first calculates the waiting time `w_2` at time `t_2`.
```
jm-clius: The ordering here doesn't quite make sense - it says we first compute `w_2` and _then_ do the computations to compute `w_2`. Better state:

```
When a new ticket request arrives, the registrar calculates `w_2` in the following way:
```

docs/ift-ts/raw/logos-service-discovery.md
```diff
@@ -1325,38 +1325,59 @@ repeatedly request new tickets hoping for better waiting times,
...
 The total waiting time will respect the lower bound if lower bound is enforced on these.
-These two sets have a bounded size as number of `ads` present in the `ad_cache`
-at a time is bounded by the cache capacity `C`.
+Service-level entries are bounded by the number of distinct `service_id_hash` values
+for which non-expired `WAIT` tickets have been issued.
+IP-level entries are bounded by the number of distinct `IPs`
+appearing in advertisements associated with non-expired `WAIT` tickets.
 
-**How SHOULD lower bound be calculated for service IDs:**
+**How SHOULD lower bound be calculated**
```
jm-clius:
```suggestion
**Lower bounds SHOULD be calculated as follows:**
```

docs/anoncomms/raw/logos-service-discovery.md
```diff
@@ -1325,38 +1325,59 @@ repeatedly request new tickets hoping for better waiting times,
...
 The total waiting time will respect the lower bound if lower bound is enforced on these.
-These two sets have a bounded size as number of `ads` present in the `ad_cache`
-at a time is bounded by the cache capacity `C`.
+Service-level entries are bounded by the number of distinct `service_id_hash` values
+for which non-expired `WAIT` tickets have been issued.
+IP-level entries are bounded by the number of distinct `IPs`
+appearing in advertisements associated with non-expired `WAIT` tickets.
 
-**How SHOULD lower bound be calculated for service IDs:**
+**How SHOULD lower bound be calculated**
```
jm-clius, line 1359: We're missing the entry conditions here (what we set bound to when a new `service_id_hash` enters teh cache). It was removed.

docs/ift-ts/raw/logos-service-discovery.md
```diff
@@ -1325,38 +1325,59 @@ repeatedly request new tickets hoping for better waiting times,
...
-the registrar calculates the service waiting time `w_s` and then applies the lower-bound rule:
+When a new ticket request arrives, the registrar first calculates the waiting time `w_2` at time `t_2`.
 
-`w_s = max(w_s, bound(service_id_hash) - timestamp(service_id_hash))`
+Then, for the corresponding `service_id_hash`, it calculates the remaining lower bound:
 
-The values `bound(service_id_hash)` and `timestamp(service_id_hash)`
-are updated whenever a new `ticket` is issued
-and the condition `w_s > (bound(service_id_hash) - timestamp(service_id_hash))`is satisfied.
+`remaining_bound = bound(service_id_hash) - (t_2 - timestamp(service_id_hash))`
```
jm-clius: Perhaps will be clearer if we first extract `t2 - timestamp(service_id_hash)` as `elapsed_time`?

jm-clius (changes requested):
Thanks! A couple of suggestions below.

docs/ift-ts/raw/logos-service-discovery.md
tinniaru3005: yup that would be better

docs/ift-ts/raw/logos-service-discovery.md
tinniaru3005: yes done. Thanks fr the suggestion

docs/ift-ts/raw/logos-service-discovery.md
tinniaru3005: yup that would be much cleaner

docs/anoncomms/raw/logos-service-discovery.md
tinniaru3005: Yup. Thanks for pointing that out

docs/anoncomms/raw/logos-service-discovery.md
```diff
@@ -1322,41 +1322,96 @@ send `ads` for random services or from diverse IPs.
...
+it first calculates the elapsed time since the bound was last updated:
+
+```text
+elapsed_time = t_2 - timestamp(service_id_hash)
+```
+
+It then calculates the remaining lower bound:
+
+```
+remaining_bound = bound(service_id_hash) - elapsed_time
```
Cofson: When `elapsed_time > bound(service_id_hash)` (an entry that has gone idle for longer than its previous bound), `remaining_bound` is negative on this line. The following `max(w_2, remaining_bound)` handles that fine, but it is the kind of thing an implementer might patch with a defensive `max(0, ...)` and silently change the semantics. A one-line note saying that `remaining_bound` MAY be negative and that the `max` with `w_2` is intentional would keep the intent explicit.

docs/anoncomms/raw/logos-service-discovery.md
```diff
@@ -1322,41 +1322,96 @@ send `ads` for random services or from diverse IPs.
...
+```
+
+The waiting time is then adjusted as:
+
+`w_2 = max(w_2, remaining_bound)`
+
+After the ticket is issued, the registrar updates the lower-bound state only if the issued waiting time is greater than the remaining lower bound:
+
+```
+if w_2 > remaining_bound:
```
Cofson: Subtle: this condition fires only when the formula's pre-clamp `w_2` already exceeded `remaining_bound`, since the line right before this block reassigns `w_2 = max(w_2, remaining_bound)`. The semantics are correct (only advance the anchor when the issued waiting time actually raises the floor), but the reader has to step through the assignments to see it. Reads more naturally with two distinct names, e.g. `w_issued = max(w_2, remaining_bound)` and then `if w_2 > remaining_bound: bound = w_issued; timestamp = t_2`.

docs/anoncomms/raw/logos-service-discovery.md
```diff
@@ -1322,41 +1322,96 @@ send `ads` for random services or from diverse IPs.
...
+  bound(service_id_hash) = w_2
+  timestamp(service_id_hash) = t_2
+```
+
+The same logic is applied for each `IP` associated with the advertisement:
```
Cofson: Small consistency thing worth pinning down: the state-keeping section above (line 1340) says `For each IP prefix in the IP tree, the registrar MUST maintain bound(IP)/timestamp(IP)` (so the keys are nodes in the prefix tree). Here the same pseudocode runs 'for each `IP` associated with the advertisement'. Is the per-IP iteration walking the prefix nodes that match the advertiser's address, or is it just the leaf IP? Different implementations will likely diverge on which set of bounds gets updated per request unless this is explicit.

Cofson (commented):
Reads cleanly after the rework, and the bound/timestamp pair is properly anchored to the invariant now. A few things worth tightening in the new lower-bound block:

docs/anoncomms/raw/logos-service-discovery.md
tinniaru3005: Good point. `remaining_bound` becoming negative when `elapsed_time > bound(...)` is intentional here, since it naturally means the previous lower bound has fully decayed. The subsequent `max(w_2, remaining_bound)` is expected to handle that case. Adding a line here.

docs/anoncomms/raw/logos-service-discovery.md
tinniaru3005: Yes, that makes sense. Thanks for the suggestion

docs/anoncomms/raw/logos-service-discovery.md
tinniaru3005: Yeah, good catch. The IP-prefix concept belongs to the IP tree logic, not the lower-bound tracking itself. Here the intent is simply to maintain lower bounds per IP.

docs/anoncomms/raw/logos-service-discovery.md
```diff
@@ -1322,41 +1322,99 @@ send `ads` for random services or from diverse IPs.
...
+
+if w_2 > remaining_bound:
+  bound(IP) = w_issued
+  timestamp(IP) = t_2
+```
+
+The lower-bound state is removed when the corresponding entry expires.
+
+This ensures that both the service-based and IP-based waiting times
+respect the lower-bound rule before the final ticket waiting time is issued.
```
jm-clius: I'm a bit confused what the final ticket "issued" waiting time will be? We've computed `w_issued` for both IP and the service_id_hash so would have ended up with two different `w_issued` values - which one would we send to the requester? Another suggestion - using the exact same symbols (`w2`, `w_issued`) etc. for both the IP and service ID formulas could be confusing as it seems like these will be the exact same values. I'd also steer clear from `_1` and `_2`, which suggests a two-step process, rather than something continuous. To disambiguate, consider something like `w(n)_ip` (nth waiting time for the IP, or perhaps "current" waiting time) and `w(n+1)_ip` (n+1th waiting time for the IP, or perhaps "new" waiting time)

jm-clius (changes requested):
Almost there. :) One more question below.

docs/anoncomms/raw/logos-service-discovery.md
tinniaru3005: Thanks, this makes sense. I agree that using the same `w_issued` for both service and IP is confusing.

What I meant is that we first compute the base waiting time, then independently enforce the lower bound for the `service_id_hash` and for each associated IP. The final ticket waiting time should be the maximum of all enforced values, so it satisfies both constraints.

I’ll update the text to use separate names.

jm-clius (approved):
Thanks!

fbarbu15 (approved):

PR #315, opened 2026-04-22
Migrate logos-messaging/specs into docs/messaging/

Cofson:
Migrates the contents of [`logos-messaging/specs`](https://github.com/logos-messaging/specs) into `docs/messaging/`.

YAML frontmatter is rewritten as a COSS pipe-table metadata header, and body links to other specs repos (`vacp2p/rfc-index`, `waku-org/specs`, `logos-messaging/specs`) and `rfc.vac.dev` are pointed at `logos-co/logos-lips`. Every spec lands at `Status: raw`; slugs will get assigned at draft promotion by the metadata validator.

## Classification

Each source file falls into one of three buckets:

- **NEW**: nothing similar exists in the destination, so it is migrated as a new raw RFC.
- **SUCCESSOR**: a newer version of an already-numbered spec; the historical numbered version stays in place.
- **SKIPPED**: the destination already holds a more developed or more recent version, so nothing is written.

### `standards/core/`

- `incentivization.md` → `docs/messaging/standards/core/incentivization.md` ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/incentivization.md))
- `mix.md` → `docs/messaging/standards/core/mix.md` (Waku Mix integration spec — distinct topic from `docs/ift-ts/raw/mix.md` which specifies the libp2p mix protocol itself) ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/mix.md))
- `relay-sharding.md` → `docs/messaging/standards/core/relay-sharding.md` ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/relay-sharding.md))
- `rendezvous.md` → `docs/messaging/standards/core/rendezvous.md` ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/rendezvous.md))
- `rln-contract.md` → `docs/messaging/standards/core/rln-contract.md` ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/rln-contract.md))
- `store-sync.md` → `docs/messaging/standards/core/store-sync.md` ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/store-sync.md))
- `sync.md` → `docs/messaging/standards/core/sync.md` ([source](https://github.com/logos-messaging/specs/blob/master/standards/core/sync.md))

### `standards/application/`

- `OpChan.md` → `docs/messaging/standards/application/opchan.md` (lowercased to match dest convention) ([source](https://github.com/logos-messaging/specs/blob/master/standards/application/OpChan.md))
[...]

PR #323, opened 2026-04-29
Add ConversationTypes Spec

jazzz:
This PR adds a new messaging specification for 'ConversationTypes'

Tasks:
- [x] ~Determine appropriate Slug value~ slugs not required any more for raw specs
- [x] Resolve linter bug
    - linter is failing due to #324


docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+| Name | ConversationTypes |
+| Slug | 9000 |
+| Status | raw |
+| Type | RFC |
+| Category | Standards Track |
+| Editor | jazzz <jazz@status.im> |
+
+## Terminology
+
+This specification uses the terminology outlined in [CHAT_DEFS](https://github.com/logos-messaging/specs/blob/master/informational/chatdefs.md) including:
```
igor-sirotin, line 14: We'll need to update the link when https://github.com/logos-co/logos-lips/pull/315 is merged

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+### Definitions
+
+A ConversationType is a specification for full-duplex communication.
+
+A Conversation is a software instance of that specification.
+
+### Scope
+
+A Conversation can be considered a processor which converts between Content and Payloads. The scope of a ConversationType defines everything inside that boundary.
+
```
igor-sirotin, line 58: If a ConversationType defines Encryption, does it mean that conversation participants are also in scope of a conversation? or a ConversationType?

I see that it's also shortly mentioned in `Conversation Rotation`:
> A new Conversation is initialized with the same membership

But conversation membership is never defined. Or am I missing something?

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+### Interface
+
+```mermaid
+sequenceDiagram
+    participant A as Client
+    participant C as Conversation
+
+    Note over A,C : Initialization
+    A ->> C: constructor(...)
+    C -->> A: ret: List<deliveryAddr>
```
igor-sirotin: Is it correct that this diagram doesn't intend to define the actual interface of `Conversation`, which all implementations must follow, but it is rather an overview, which operations can be done with a `Conversation`?

I agree that _Initialization_ should produce a `List<deliveryAddr>`, which the client should subscribe to (using Delivery Service). But here it looks like a `constructor()` would _return_ a list of DA, rather than returning an instance of `Conversation`.

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+
+Conversation Rotation is the process of migrating participants from one ConversationType to a new one. A new Conversation is initialized with the same membership and the existing Conversation is archived. This is the primary mechanism for accessing new functionality without breaking existing clients.
+
+### Protocol Naming
+
+ConversationType names are for convenience only — they carry no semantic meaning and imply no protocol relationship, compatibility, or ordering. Similarly named ConversationTypes do not imply any relationship between them. Implementors MAY use a common prefix for organizational clarity.
+
+### Self-Describing Messages
+
+Choosing self-describing protocol messages is always preferred — it reduces ambiguity and makes implementations easier to debug and extend.
```
igor-sirotin: What exactly `Self-Describing Messages` is about, is it encoding/decoding? Or is there a different meaning?

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Implementation Suggestions
+
+### Logical "Chats"
+
+App developers should maintain a logical separation between the user-facing message stream (a "chat") and the Conversation used to transport it. As ConversationTypes are immutable, the lifecycle of a "chat" may outlive the Conversation that carries it — particularly across Conversation Rotation. Applications that map a stable "ChatId" to a current ConversationId will handle rotation more gracefully.
+
+### Conversation Rotation
+
+Conversation Rotation is the process of migrating participants from one ConversationType to a new one. A new Conversation is initialized with the same membership and the existing Conversation is archived. This is the primary mechanism for accessing new functionality without breaking existing clients.
```
igor-sirotin: Should this spec define how to perform a Conversation Rotation?

Knowing that we're doing it in a decentralized way, how should other clients know when to start using a new ConversationType, including when to subscribe to new delivery addresses and unsubscribe from the old ones.

If this is to be solved by applications, maybe we could give suggestions in this spec?

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+A single Content message MAY result in multiple Payloads, each with its own delivery address. This allows a ConversationType to fan out messages across multiple delivery addresses when required by its design.
+
+An inbound Payload will either produce a single Content or nothing. The None case is intentional — not all Payloads are intended to surface content to the application. Some may be protocol-level messages handled internally by the Conversation.
+
+**Requirements:**
+- All Conversation instances MUST provide the following conversions:
+  - `Content -> list(DeliveryAddress, Payload)`
+  - `Payload -> Content | None`
+
+### Immutability
```
igor-sirotin, line 146: I wonder of apps binary size consequences. 

I assume that:
- new ConversationTypes are introduced rarely
- and rotations will probably reuse internal parts
- and in general we're probably talking about <1Mb of binary size per conversation type.

But if all of these are wrong, it might become an issue. 

I guess that after rotations, old ConversationTypes would have to retain in application binaries forever. Or is it possible to give any recommendations for applications on how to drop old ConversationTypes, not to carry useless code?

igor-sirotin (commented):
Love it!

docs/messaging/raw/conversationtypes.md
jazzz: Good point. 

Yes, membership is state managed by a conversation.
The conversationType would define rules for membership, and the mechanism that allows for membership changes.

E.g AddMember would be a specific Frame type defined by a given ConversationType

I'll update

docs/messaging/raw/conversationtypes.md
jazzz: > Is it correct that this diagram doesn't intend to define the actual interface of Conversation, which all implementations must follow

Correct. The actual interface is an implementation detail and not required for interoperation between applications.  
 
 The real practical limitation is:
 
 > A Conversation instance MUST define a static set of `delivery addresses` to subscribe to during initialization.
 
 Allowing a dynamic set of subscriptions complicates implementation and coordination of clients - For now easiest to define that these are needed to be determinable at the moment of initialization.

jazzz: The goal of the diagram was to show a highlevel flow, but perhaps thats confusing more than its helping. 

The "client" is largely irrelevant here: ConversationTypes need to provide a small subset of functionality, probably best just to focus on that.

docs/messaging/raw/conversationtypes.md
jazzz: > new ConversationTypes are introduced rarely

It depends on your definitions, but I see new ConversationTypes being added regularly. The intention is that one does not have to wait to deploy breaking changes. Rotate to a new ConversationType, and continue with developement.

> and rotations will probably reuse internal parts

Yes. The vast majority of ConversationTypes will be small changes, to existing ConversationTypes.

> and in general we're probably talking about <1Mb of binary size per conversation type.

That likely depends heavily on language choice and compiler settings - far outside the scope here. That feels right though for what that is worth.

jazzz: > I guess that after rotations, old ConversationTypes would have to retain in application binaries forever.

Not true. 

SDK implementations, can support what ever ConversationTypes they want, including dropping support for older types that are not in use.

The ConversationTypes that an SDK implementation needs to support is directly related to which are in use by the network.

> ...on how to drop old ConversationTypes, not to carry useless code?

Absolutely. The Specification that pairs with this, which covers the items not listed here (Envelopes, Type discrimination, capability discovery) will touch on this. 

Support for ConversationTypes will likely follow a 'sliding window' where new types are added, and others fallout of support after some time. The tension here is fragmentation, and guidance for SDK developers will help manage that.
 
Recommendations include: 
- New chats SHOULD be initialized using the newest support edConversationTypes
- existing chats SHOULD rotate to new versions, when the participants support it.

These two mechanisms mean that the conversationTypes in use will follow a longtailed normal distribution skewed towards the most recent Types.

jazzz: > Or is it possible to give any recommendations for applications...

One thing to note: I'm purposely avoiding mentioning "applications" in this work. I use implementation to refer to the "sdks" and "libraries" that implement these ConversationTypes (TODO: Define implemention in spec). 

How and What SDK's expose to developers is an implementation choice, and beyond support for ConversationTypes has no bearing on interop. These specs are aimed at SDK/Library developers - they choose how to best serve their developers.

docs/messaging/raw/conversationtypes.md
jazzz: Self describing messages is a protocol design approach. The goal is to decouple the wire formats from the software that generated it.

This creates a more temporally stable, and version resistant protocol.

Currently this is not that helpful to implementors - I'll drop this section for now, and revisit later.

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+A ConversationType is a fixed contract — once published it cannot be changed. This eliminates compatibility mismatches between clients: any two clients that support the same ConversationType can always interoperate, regardless of their implementation or version.
+
+When new functionality is needed, a new ConversationType is defined. Existing ConversationTypes remain valid indefinitely.
+
+Immutability applies to frame definitions and their semantics — not to the data carried within frames.
+
+Because ConversationTypes are immutable, implementors do not need to manage breaking schema changes. Any additions to stored state will be additive, removing the need for complex migrations, `semver`, or version tracking.
+
+**Requirements:**
+- A ConversationType MUST NOT be modified after publication.
```
osmaczko, line 159: Should these specs address how new `ConversationType`s will be published?

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Overview
+
+This specification takes a different approach to the versioning problem. Rather than versioning protocols, we make them immutable. A ConversationType is a fixed contract — it never changes. When new functionality is needed, a new ConversationType is defined and participants rotate to it. This makes upgrades cheap, explicit, and coordination-free: clients that support a ConversationType can always communicate, and adding new capabilities never breaks existing ones.
+
+ConversationTypes are the mechanism that makes this possible — a common framework for defining messaging protocols that is modular, immutable, and interoperable by design. ConversationTypes define protocol messages and their serialization, including message structures, encryption mechanisms, and encoding.
+
+In this model, clients are expected to support multiple ConversationTypes simultaneously. Compatibility between clients is determined not by client version, but by the set of ConversationTypes each client supports.
```
osmaczko: > Compatibility between clients is determined not by client version, but by the set of ConversationTypes each client supports.

I don’t think this is entirely accurate. Since the client defines the `Content`, it may introduce breaking changes between versions regardless of the supported `ConversationTypes`.

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,190 @@
...
+
+**Requirements:**
+- ConversationIds MUST be 128bits long.
+- For every client, a ConversationId MUST reference one and only one Conversation.
+
+### Delivery Subscriptions
+
+Delivery addresses define where a Conversation expects to receive inbound Payloads from the DS. They are defined by the ConversationType rather than the DS for two reasons:
+
+First, delivery addresses have direct privacy implications — how messages are grouped and addressed can leak metadata about participants and group membership. The ConversationType is best positioned to reason about this, as it has full knowledge of the message structure and encryption scheme.
```
osmaczko: Sounds valid. Just raising a concern: some clients might want to influence the delivery address. For instance, in Status there was a feature called “static sharding,” where a protected shard was set up for a given Community. It was not the best design from a privacy standpoint, but this illustrates that future use cases may require clients to influence the delivery address.

osmaczko (commented):
Thanks for these specs, they give very good insight into your vision.

In general, I like the idea of immutable ConversationTypes, as it removes the compatibility overhead from the most critical parts of chat: segmentation, reliability, encryption, encoding, and addressing. This gives interoperability at the conversation level and shifts all the compatibility burden to the client (aka chat), which is fine.

docs/messaging/raw/conversationtypes.md
jazzz: > Should this spec define how to perform a Conversation Rotation?

I don't think so. Conversation rotation is not a required function of all conversationTypes. There is no "Function" that needs to be called, no interoperability concerns from the viewpoint of all conversationTypes. This spec is the contract between a client and a ConversationType. 

How conversations like GroupV1 rotate to a new conversation, is defined in GroupV1. PrivateV2 might do it differently. 

This is not an attempt at being "super modular"; its an artifact that the logic for when to rotate, under what circumstances and how to determine where to rotate too are all ConversationType specific.

> If this is to be solved by applications...

Not solved by Application, its solved by specific ConversationTypes. Rotation is completely defined within the Protocol layer and should be opaque to applications. 

>  maybe we could give suggestions 

Absolutely. Once the first ConversationType implements it there will be a recipe that can be followed.

docs/messaging/raw/conversationtypes.md
jazzz: Thats a fair point for a follow up PR. 

The short answer is that a permissionless registry similar to [ContentTypes](https://github.com/logos-messaging/specs/blob/master/standards/application/contentframe.md) will be needed. 

This seems like something we need to address once capability discovery is being worked on.

docs/messaging/raw/conversationtypes.md
jazzz: > Since the client defines the Content

Clients SHOULD to be Content agnostic - thought not really a requirement for interoperation. 

Supported ContentTypes SHOULD be determined by the applications, and they can theoretically transmit arbitrary bytes across the wire. In practice this would have low interoperability because no one else can understand the data format, however thats up to developers to choose. If you specifically only want to send messages between other instances of your application then there is no need to use the same types as anyone else. 

To help developers interop between other applications the [ContentTypes specification](https://github.com/logos-messaging/specs/blob/master/standards/application/contentframe.md) provides developers with a tool to coordinate between themselves. 

Developers are encouraged to support a common set of types, and share their custom types with the community - But currently no plan. 

Content type innovation comes from application developers who are trying to bring features to their users - its critical that they remain autonomy over this. That said fragmentation is a real issue, having a community maintained library that developers can use will help keep applications compatible.

docs/messaging/raw/conversationtypes.md
jazzz: > some clients might want to influence the delivery address
> For instance, in Status there..

Are you using a different definition of [Client](https://github.com/logos-messaging/specs/blob/master/informational/chatdefs.md#software-entities)? Or did you mean Application?

jazzz: > For instance, in Status there was a feature called “static sharding,” where a protected shard was set up for a given Community.

My hot take: "Applications" (Status included) shouldn't be concerning themselves with such concepts. 

Delivery addresses also have impacts on bandwidth utilization and interoperability. A client making decisions here would not necessarily be reflected in other applications. 

If developers have usecases where having fine grained control over internal protocol operation is  critical; Then they ought to be defining their own ConversationTypes. 

Applications, Clients, and ConversationTypes are perspectives. The layers concretely define "where" a problem should be solved, not "who" should be solving it. A developer could contribute to an app, an SDK implementation, and develop ConversationTypes.  

Each layer has its own "autonomy" and "iterop" domains; Areas its free to make decisions for its users, and areas it needs to ensure compatibility or cause fragment

| Layer | Autonomy Domain | Autonomy Surface | Interop Domain | Interop Surface |
|---|---|---|---|---|
| **Applications** | Their Users | UI/UX, Integration, <br>Message retention | A Users Contacts | Content Types |
| **Clients** |  Their Apps | API,<br> Delivery,<br>State Storage, <br>Device Support | Network Clients | Envelopes,<br> Routing, <br>ConversationTypes |
| **Conversation<br>Types** | Protocol Features | Privacy Properties, <br>Encryption, <br>Membership,<br> Admin rules, <br>Protocol features | Convo Members | Frames, <br>State Updates |

docs/messaging/raw/conversationtypes.md
osmaczko: What I meant is that ~~Clients~~Applications? might decide to introduce breaking change in their content type on their own will, breaking older clients, regardless of supported set of ConversationTypes.

jazzz: > I don’t think this is entirely accurate.

To address the original post; The statement talks exclusively about Clients, and clients SHOULD/MUST? be content agnostic. So I think its accurate unless I am missing something. Highlighted in this [table](https://github.com/logos-co/logos-lips/pull/323#discussion_r3183381736) client interop does not depend on content, thats a application responsibility.


> What I meant is that Applications? might decide to introduce breaking change in their content type on their own will, breaking older clients

Correct. With the caveat operationally the clients remain interoperable; the outcome is that perhaps not all applications will be able to display the message.  

This is a well establish problem between extensibility and fragmentation. 

- Giving developers infinite choice results results in fragmentation of the network; where not every application can understand each other.
- Giving developers strict guard rails, stops them from being able to solve the problems there users are having.

jazzz: These problems exist regardless if we let go of web2 based deployment assumptions.

I hold as truth:
- Apps can force their Users to upgrade, as much as SDK developers can force applications to update.
- Users of the same App are not guaranteed to be compatible as they may be running different versions. 

Even with centralized/protocol mandated content types, older applications will not be able to understand message types it does not have definitions for. 

Sidestepping Theory and math;

As described in the future logos chat spec, rather than long release cycles and slow feature adoption (aka: code soaking/read-before-write deployment)l Why not let developers choose what is best for themselves.

Theres no requirement that there is only one `VoiceMessage` type in the ecosystem. Developers are motivated to support types which:
- Provide the functionality their users want to send,
- Support message types that their users are receiving.

Either they are motivated to collaborate with other developers for a mutually beneficial experience  or they are not, and they will go their own path.

docs/messaging/raw/conversationtypes.md
osmaczko: Thanks for explanations, especially for a table with comparison. I indeed had wrong mental model of the Client.

> If developers have usecases where having fine grained control over internal protocol operation is critical; Then they ought to be defining their own ConversationTypes.

Indeed, makes sense :+1:

docs/messaging/raw/conversationtypes.md
osmaczko: image

docs/messaging/raw/conversationtypes.md
igor-sirotin: TIL the word `temporally` 😄

docs/messaging/raw/conversationtypes.md
jazzz: If Client is a confusing term, I'm open to a new name. 

The name should Convey:
- its network aware
- its platform specific
- its role is to connect to other instances
- Applications delegate to it for operation.

docs/messaging/raw/conversationtypes.md
jazzz: Changed my mind. Relabeled the section to "High Level Flow" and then simplified the constructor and left it to implementors.

docs/messaging/raw/conversationtypes.md
jazzz: removed

jazzz:
@igor-sirotin @osmaczko 

I'm not seeing any more open comments that require updated. Is there any blocking this from approval?

This is being submitted as a 'RAW' spec so there is no real commitment required. This just represents our current 'best approach' while we refine and implement.

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,226 @@
...
+- All Conversation instances MUST provide the following conversions:
+  - `Content -> list(DeliveryAddress, Payload)`
+  - `Payload -> Content | None`
+
+### Immutability
+
+A ConversationType is a fixed contract — once published it cannot be changed. 
+This eliminates compatibility mismatches between clients: any two clients that support the same ConversationType can always interoperate, regardless of their implementation or version.
+
+When new functionality is needed, a new ConversationType is defined. Existing ConversationTypes remain valid indefinitely.
```
kaichaosun, line 151: What if there are security patches which needs immediate updates for existing ConversationType in a short time window? In this case, release a new conversation type with the patch will result the existing clients in risk.

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,226 @@
...
+
+Immutability applies to frame definitions and their semantics — not to the data carried within frames.
+
+Because ConversationTypes are immutable, implementors do not need to manage breaking schema changes.
```
kaichaosun, line 155: Does it mean every new conversation type will have its own tables in storage layer like sqlite? For example, if we add a small feature in PrivateV1, which result in PrivateV2 with a new added column, do we have fresh new tables or reusing existing ones?

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,226 @@
...
+When new functionality is needed, a new ConversationType is defined. Existing ConversationTypes remain valid indefinitely.
+
+Immutability applies to frame definitions and their semantics — not to the data carried within frames.
+
+Because ConversationTypes are immutable, implementors do not need to manage breaking schema changes. 
+Any additions to stored state will be additive, removing the need for complex migrations, `semver`, or version tracking.
+
+**Requirements:**
+- A ConversationType MUST NOT be modified after publication.
+- Any change that affects interoperability MUST be defined as a new ConversationType.
```
kaichaosun, line 160: A logos-messaging content address uri `xyz/privatev1` could encode variants or dynamic information and not break interoperability, but may breaks immutability. 

Actually I'm confused with immutability by looking existing definition. How is such immutability been enforced? For example, IPFS content is immutable, every change to the content result into a new hash, smart contract is immutable with similar mechanism.

osmaczko (approved):

docs/messaging/raw/conversationtypes.md
osmaczko: I think this might just be me being slow to adapt.

Nevertheless, "Client" is a fairly ambiguous and broadly used term, and in the context of layered code it tends to collapse into noise. Consider the following sentence: "Application is a client of Client, Client is a client of the Conversations".

I have a few proposals, but if none of them fits, I am fine staying with "Client". My brain just needs an adaptation period.

"Applications use _..._ to participate in Conversations."

1) **ChatEngine**: _Application -> ChatEngine -> Conversations_
2) **ChatRuntime**: _Application -> ChatRuntime -> Conversations_
3) **Messenger**: _Application -> Messenger -> Conversations_
4) **Chat**: _Application -> Chat -> Conversations_

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,226 @@
...
+    C -->> A : ret: Content or None
+```
+
+## Assumptions
+
+### Delivery Service
+
+This specification assumes the existence of a Delivery Service responsible for routing payloads between clients. A ConversationType is always defined in the context of a DS with the following properties:
+
+- A DS operates as a PubSub — it supports Publish and Subscribe functionality.
```
Cofson, line 91: `DS` is introduced inline here as the abbreviation for Delivery Service and recurs throughout the rest of this section. It is not in the Terminology block at the top, so a reader who lands here first has to backtrack. Either add `Delivery Service (DS)` to Terminology, or spell it out on first use in line 89 ('A Delivery Service (DS) operates as a PubSub...').

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,226 @@
...
+They sit between the raw Payload bytes and the Content exposed to the application — a Payload is decoded into a Frame, and either handled internally by the Conversation or converted to Content.
+
+Each ConversationType defines its own set of Frames. 
+Multiple ConversationTypes may reuse the same Frame definitions but there is no requirement to do so.
+
+All Frames fall into one of three states when processed by a Conversation. 
+Distinguishing between these states improves observability — a malformed Frame and an unrecognized Frame are different problems and should not be conflated:
+- **Valid** — the Frame is well-formed and recognized.
+- **Invalid** — the Frame is malformed or fails validation.
+- **Unsupported** — the Frame is unrecognized by this Conversation instance.
```
Cofson, line 174: The three Frame states are well defined and the spec requires the ConversationType to distinguish between them, but it does not say what the Conversation should actually do with each. Drop, propagate, surface to the application, count toward a peer-scoring signal? Even leaving the choice implementation-defined, an explicit statement would help two implementations agree on observable behaviour for Unsupported frames in particular.

docs/messaging/raw/conversationtypes.md
```diff
@@ -0,0 +1,226 @@
...
+### Conversation Rotation
+
+Conversation Rotation is the process of migrating participants from one ConversationType to a new one. 
+A new Conversation is initialized with the same membership and the existing Conversation is archived. 
+This is the primary mechanism for accessing new functionality without breaking existing clients.
+
+### Protocol Naming
+
+ConversationType names are for convenience only — they carry no semantic meaning and imply no protocol relationship, compatibility, or ordering. 
+`Similarly named ConversationTypes do not imply any relationship between them. Implementors MAY use a common prefix for organizational clarity.
```
Cofson:
```suggestion
Similarly named ConversationTypes do not imply any relationship between them. Implementors MAY use a common prefix for organizational clarity.
```
There's a stray opening backtick at the start of this line that is never closed, so everything from here through the next backtick (in the Security Considerations TODO area) renders as inline code in mdbook.

Cofson (commented):
Reads well overall. The immutability framing keeps the spec compact and the Frame model is clean. A few clarifications worth pinning down on the way to draft:

docs/messaging/raw/conversationtypes.md
igor-sirotin: > in Status there was a feature called “static sharding,” where a protected shard was set up for a given Community

@osmaczko off-topic, but this feature is dead, isn't it?

igor-sirotin (approved):

docs/messaging/raw/conversationtypes.md
jazzz: >  In this case, release a new conversation type with the patch will result the existing clients in risk.

This is true. But the problem is not unique to a this approach - its the same if you wanted to release a new version of the protocol. Deploying security patches in a decentralized protocol, is not a fast process. No one can force developers to update, and developers can't really force users to update.

The solutions come from, a security patch notification system, maintaining a great relationship between application developers - all independent of conversationType immutability.

In these cases, having the ability to deploy security patches without breaking compatibility helps resolve the problem more quickly, with less headaches.

docs/messaging/raw/conversationtypes.md
jazzz: > Does it mean every new conversation type will have its own tables in storage layer like sqlite?
> do we have fresh new tables or reusing existing ones?

Depends on the ConversationType, but in most cases; New ConversationTypes -> New storage requirements -> New tables

docs/messaging/raw/conversationtypes.md
jazzz: > How is such immutability been enforced?

I think the the part that is missing is:

> [A ConversationType is a specification for full-duplex communication. ](https://github.com/logos-co/logos-lips/blob/3f6775b7932728e02db47ad09b74029f29d5800f/docs/messaging/raw/conversationtypes.md?plain=1#L50)

Immutability applies to the protocol definition (ConversationType) not to instances of it (Conversations). 

Conversations can have `dynamic` fields  such as `delivery_address`. However the process for deriving those fields cannot change once published. Every change results in some clients believing two different truths - this cannot happen.

docs/messaging/raw/conversationtypes.md
jazzz: Great call out. 

Updated first use to read "...assumes the existence of a Delivery Service **(DS)** responsible for routing payloads..."

Cofson:
@igor-sirotin @jazzz  FYI I forced push some changes onto the original PR (#315 ). This one will probably need rebasing. Ignore if not - just wanted to let you know.

kaichaosun (approved):

docs/messaging/raw/conversationtypes.md
osmaczko: > > in Status there was a feature called “static sharding,” where a protected shard was set up for a given Community
> 
> @osmaczko off-topic, but this feature is dead, isn't it?

It is :skull:

docs/messaging/raw/conversationtypes.md
jazzz: This is a great point. I'll circle back to it after some thought, and the other companion specifications land.

PR #327, opened 2026-05-04
Stake-Weighted Mix RLN DoS Protection spec

AkshayaMani:
This PR adds a registration policy extension for [RLN Per-Hop DoS Protection](https://github.com/logos-co/logos-lips/blob/master/docs/ift-ts/raw/mix-spam-protection-rln.md). It replaces the flat per-node rate limit `R_base` with a linear stake-to-rate mapping enforced via the RLN-Diff circuit, raising the economic cost of exploiting the rate amplification gap. The mapping is Sybil-resistant by construction and requires no changes to the [RLN-v2 ZK circuit](https://github.com/vacp2p/rfc-index/blob/dabc31786b4a4ca704ebcd1105239faff7ac2b47/vac/raw/rln-v2.md) — rate differentiation is enforced entirely by the membership registry.

### Contents

- **§1 Introduction:** Motivates replacing the flat rate limit with a stake-proportional mapping
- **§2 Terminology:** Introduces `S_unit`, `R_base`, `R_min`, `R_max`, floor-stake, Membership, and Sybil-resistant
- **§3 Background:** Defines the rate amplification gap in [RLN Per-Hop DoS Protection](https://github.com/logos-co/logos-lips/blob/master/docs/ift-ts/raw/mix-spam-protection-rln.md) and provides background on RLN-Diff (registration, rate limit, double-signalling and slashing)
- **§4 Approach:** Specifies the linear stake-to-rate mapping, registration mechanics, packet sending and verification, and system parameters (`S_unit`, `R_min`, `R_max = f × R_base`, `f`)
- **§5 Security and Privacy Considerations:** Analyzes Sybil-resistance, residual rate amplification gap, and registry-layer observability
- **§6 Out of Scope:** Items deferred to deployment policy (_e.g.,_ voluntary deregistration, migration procedures)
- **§7 Future Work:** Registry-layer privacy via shielded registry, dynamic stake top-up, and cover traffic under non-uniform rates


docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+
+### 5.2 Residual Rate Amplification Gap
+
+The stake-to-rate mapping raises the economic cost of exploiting the rate amplification gap defined in [Section 3.1](#31-rate-amplification-gap), but does not eliminate the gap.
+A malicious node that wants a high message rate must commit proportionally more stake, which is subject to slashing on detection.
+
+The gap is structural: the [Mix Protocol](./mix.md) unlinkability guarantees make forwarding and origination indistinguishable, so any forwarding rate allowance is simultaneously an origination budget.
+This is an explicitly acknowledged limitation of per-hop RLN regardless of the rate mapping function.
+
+### 5.3 Registry-Layer Observability
```
chaitanyaprem: wondering that by allowing any rates between R_min and R_max, an observer of node's traffic can connect the node's stake and in-turn link the wallet address used for registration. 
Would it make it sense to have a set of pre-defined rates maybe 5 that all users can use/subscribe to which would increase the anonymity set for an observer?

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+
+## 3. Background
+
+This section provides background on the rate amplification gap in [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md), and RLN-Diff.
+
+### 3.1 Rate Amplification Gap
+
+[RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md) enforces a flat rate limit `R_base` per node per epoch on outgoing packets. Two properties combine to make this exploitable:
+
+- `R_base` must be set high enough to accommodate the forwarding load of high-capacity relay nodes. As a result, all nodes receive the same high `R_base` regardless of stake.
```
chaitanyaprem: Not sure which section which should mention the limitation, but using varying rate-limits would need modification of mix node path selection logic from uniform to stake-weighted. Otherwise there is a higher possibility of forwarding traffic getting dropped as low stake nodes that get selected more often would keep hitting rate limits.

chaitanyaprem (commented):
Will do another pass, but wanted to highlight 2 points.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+| Editor | Akshaya Mani <akshaya@status.im> |
+| Contributors |                              |
+
+<!-- timeline:start -->
+<!-- timeline:end -->
+
+
+## Abstract
+
+This document specifies a registration policy extension for the [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md) specification, introducing stake-proportional rate limits for mix nodes.
```
chaitanyaprem: i think we may have to rename mix-spam-protection-rln.md to mix-dos-protection-rln.md and update refs. 
Need not be done as part of this PR, but would be good to do it.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+Under this extension, each node's rate limit is computed as a linear function of its committed stake and enforced via the RLN-Diff circuit defined in [RLN-v2](https://github.com/vacp2p/rfc-index/blob/dabc31786b4a4ca704ebcd1105239faff7ac2b47/vac/raw/rln-v2.md).
+The mapping is Sybil-resistant, requires no circuit changes, and enforces rate differentiation entirely through the membership registry.
+
+## 1. Introduction
+
+The [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md) assigns a flat rate limit to all mix nodes that meet a minimum stake requirement. To accommodate high-capacity relay nodes, this limit must be set high&mdash;making the same rate headroom available to any minimally-staked attacker. This structural limitation is referred to as the rate amplification gap, defined in [Section 3.1](#31-rate-amplification-gap).
```
chaitanyaprem: Nit: the dash is rendering not great, maybe some spacing needed?
Image

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
chaitanyaprem: Maybe i will do it as part of changing the original spec to mention that RLNv2 is being used.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
chaitanyaprem: especially after adding cover-traffic (whichever strategy we use) it would become easier for an observer to determine a node's rate and in-turn map to stake and probably derive the address by looking at history of transactions.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+## 3. Background
+
+This section provides background on the rate amplification gap in [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md), and RLN-Diff.
+
+### 3.1 Rate Amplification Gap
+
+[RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md) enforces a flat rate limit `R_base` per node per epoch on outgoing packets. Two properties combine to make this exploitable:
+
+- `R_base` must be set high enough to accommodate the forwarding load of high-capacity relay nodes. As a result, all nodes receive the same high `R_base` regardless of stake.
+- The [Mix Protocol](./mix.md)'s unlinkability guarantees make forwarded and originated packets cryptographically indistinguishable by design, so any node can use its `R_base` allowance entirely for origination.
```
chaitanyaprem: Maybe we can elaborate that a node can choose to drop forward traffic and only use its rate limit for origination. 
This can happen even with stake based rate limits right. 
Whatever rate-limit assignment we choose, this is still bound to happen. 

One way i can see this addressed is to use some sort of reputation which would identify that this node is not participating in the network rather only using it. Then there must also be a mechanism to again penalize the node based on this reputation.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+2. Submits `id_commitment` and a `user_message_limit` to the membership registry.
+
+The registry computes the Merkle leaf:
+
+```text
+rate_commitment = Poseidon(id_commitment, user_message_limit)
+```
+
+and writes `rate_commitment` to the RLN membership Merkle tree.
+`user_message_limit` is observable to any party that can read the registry.
```
chaitanyaprem, line 129: but they wont be able to tie it to a specific wallet address that did the registration right, or would it be possible by just tracing blockchain history?

Maybe logos-blockchain/LEZ may help here where-in any party observing this might not be able to deduce the address used for registering a certain leaf.
cc @adklempner any thoughts?

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+Raising `R_base` to accommodate higher-capacity relays widens this budget proportionally. This structural limitation is referred to as the rate amplification gap.
+
+### 3.2 RLN-Diff
+
+[Rate Limiting Nullifiers (RLN)](https://github.com/vacp2p/rfc-index/blob/dabc31786b4a4ca704ebcd1105239faff7ac2b47/vac/raw/rln-v2.md) is a zero-knowledge construct that allows message rate limits to be set and cryptographically enforced for members of a group, without revealing their identity.
+To participate, senders must register for a membership, typically backed by stake, to generate valid proofs.
+If a sender exceeds the defined rate limit within a given epoch, the protocol allows for the reconstruction of their secret and subsequent slashing of their stake.
+This specification uses the RLN-Diff variant, which supports per-member rate limits set individually at registration.
+
+**Registration.**
```
chaitanyaprem: Nit: maybe not use a dot at the end as this is a heading? it is getting rendered funnily. 

Image

Similarly for next sub sections.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+
+and writes `rate_commitment` to the RLN membership Merkle tree.
+`user_message_limit` is observable to any party that can read the registry.
+
+All members of a group share a single Merkle tree.
+The Merkle root compactly represents the current group state and changes whenever a member joins or is slashed; a proof must reference a recent root to be considered valid.
+
+**Rate limit.**
+Each member's rate is bounded by the `user_message_limit` set at registration.
+To send a packet, a member selects a `message_id` in the range `[1, user_message_limit]` that has not been used in the current epoch, and generates a ZK proof that:
```
chaitanyaprem: the below flow should be part of zerokit/RLN already right, maybe better to reference it, rather than describing it again?

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
chaitanyaprem: LEZ/Logos-blockchain may help here.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+- Dynamic rate adjustment without re-registration
+- Reputation-based rate multipliers
+- Stake token selection and blockchain infrastructure
+- Detailed migration procedures and tooling for networks upgrading from flat-rate RLN
+- Voluntary deregistration mechanisms (a [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md) responsibility)
+- Membership registry implementation (smart contract, coordination layer, or other)
+
+## 7. Future Work
+
+- **Registry-layer privacy**: The registration publicly reveals the node's effective rate. A shielded membership registry&mdash;where the stake-to-identity link is concealed via ZK&mdash;would close this gap. One approach is multi-identity registration: a node registers multiple unit-rate identities, hiding its total rate from registry observers. However, slashing requires linking all identities to the committed stake: double-signalling on any identity must make the entire stake slashable. Achieving such an unlinkable registration with linked slashing requires circuit-level changes not available in current RLN-v2.
```
chaitanyaprem: I think LEZ already supports this, will have to check with @adklempner as he is working on RLN on LEZ/logos-blockchain

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+`user_message_limit` is observable at the registry layer but is not revealed by proofs; verifiers learn only that the sender has not exceeded their registered limit.
+
+All per-hop verification and slashing logic are as defined in [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md).
+
+### 4.4 System Parameters
+
+| Parameter | Description |
+| --- | --- |
+| `S_unit` | Stake required per message per epoch. Deployment-defined. |
+| `R_min` | Minimum rate. Nodes with stake `S < R_min × S_unit` MUST be rejected. |
```
Cofson: `R_min` is the minimum rate any registered node can have, and `R_base` is the existing flat per-node rate from the parent spec. Their relationship is not constrained anywhere. If `R_min > R_base`, a floor-stake node would receive a higher rate than a node under the parent spec's flat rate, which is probably not the intent (the whole point of stake-weighting is to retain `R_base` as a baseline and only grant more rate for more stake). Worth either an explicit `R_min ≤ R_base` constraint, or a note explaining why no constraint is needed.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+The membership registry MUST enforce the following at the time of registration:
+
+1. Verify that `S ≥ R_min × S_unit`. Reject registrations below floor-stake.
+2. Compute `user_message_limit = min( floor( S / S_unit ), R_max )`.
+3. Complete registration as described in [Section 3.2](#32-rln-diff) using the computed `user_message_limit`.
+4. Lock the stake for the duration of membership.
+   Stake MUST NOT be withdrawable while membership is active.
+
+A node that wishes to increase its `user_message_limit` MUST deregister and re-register with the new stake amount.
+Partial stake top-ups that would change `user_message_limit` without re-registration MUST NOT be accepted, as the existing Merkle leaf encodes the previous rate.
```
Cofson: This rules out partial top-ups that would change `user_message_limit` without re-registration. What about top-ups that don't cross a floor bucket (e.g. 250 → 275 stake when `S_unit = 100`, leaving `floor(S/S_unit) = 2` unchanged)? Are those allowed and silently accepted by the registry, or also disallowed? The current wording is ambiguous and an implementer would have to guess. Easy fix is to require re-registration for any stake change, or to explicitly permit (and define the semantics of) no-op top-ups.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,245 @@
...
+| `R_max` | `f × R_base`. Maximum rate regardless of stake. |
+| `f` | Deployment multiplier `f ≥ 1`. Controls the maximum rate relative to the base rate. |
+
+`S_unit` SHOULD be set such that a floor-stake node generating only mandatory cover traffic at the rate specified in [Mix Cover Traffic](./mix-cover-traffic.md) can sustain operation at `R_min`.
+
+`f` MUST be set such that a single node operating at `R_max` cannot individually saturate the network's forwarding capacity.
+`f = 10` is a reasonable starting point: it allows a high-stake operator to handle `10×` the base forwarding load without dominating the network.
+Values above `f = 50` SHOULD be avoided without careful analysis of the deployment's expected node count and forwarding load distribution.
+
+`S_unit`, `R_min`, `R_max`, and `f` MUST be published in a deployment configuration accessible to all participants before the network accepts registrations.
```
Cofson: The required-published list (`S_unit`, `R_min`, `R_max`, and `f`) omits `R_base`, but `R_max` is defined as `f × R_base` (lines 44 and 169). Without `R_base` published, participants cannot independently verify that the published `R_max` is consistent with the deployment's `f`. Either add `R_base` to the list here, or note explicitly that `R_base` is inherited from [RLN Per-Hop DoS Protection](./mix-spam-protection-rln.md) and is that parent spec's responsibility to publish.

Cofson (commented):
Tight spec, and the rationale for picking linear over sub-linear or super-linear mapping is well argued. A few small technical gaps worth pinning down before draft:

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Good catch — added the constraint `1 ≤ R_min ≤ R_base` in §2 and §4.4.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Tightened to disallow stake changes while membership is active — any change requires deregistration and re-registration. This avoids registry state ambiguity around no-op top-ups and keeps slashable amount fixed at registration.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Added `R_base` to the published list.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: > Maybe we can elaborate that a node can choose to drop forward traffic and only use its rate limit for origination.

§3.1 and §5.2 already acknowledges this — "so any node can use its `R_base` allowance entirely for origination." and "any forwarding rate allowance is simultaneously an origination budget"

> One way i can see this addressed is to use some sort of reputation which would identify that this node is not participating in the network rather only using it. Then there must also be a mechanism to again penalize the node based on this reputation.

Identifying nodes that are not forwarding would require distinguishing originated from forwarded packets on a per-node basis, which is precisely what Mix Protocol's unlinkability prevents. Building such a reputation system would conflict with Mix's core anonymity guarantee. §6 also already places reputation-based mechanisms out of scope.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: The path selection change itself is a Mix Protocol concern, so left as a future work bullet (§7).

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Added spaces around `&mdash;` throughout the spec for cleaner rendering.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Converted the run-in headings to **Heading**: format

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Trimmed §3.2 to keep only the parts §4 and §5 actually depend on.

AkshayaMani: Trimmed §3.2 to keep only the parts §4 and §5 actually depend on.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: > i think we may have to rename mix-spam-protection-rln.md to mix-dos-protection-rln.md and update refs.

Done.

Cofson (approved):

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Good point! Addressed by introducing a stake tier size parameter to quantize the linear mapping (to ensure Sybil resistance holds while still providing anonymity sets per tier).

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Yes — on-chain registries would allow this via blockchain history tracing. §5.3 now describes the attack chain (its dependence on the registration mechanism); §7 references shielded registry (LEZ-style) as the complete fix.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,243 @@
...
+
+## Abstract
+
+This document specifies a registration policy extension for the [RLN Per-Hop DoS Protection](./mix-dos-protection-rln.md) specification, introducing stake-proportional rate limits for mix nodes.
+Under this extension, each node's rate limit is proportional to its committed stake and enforced via the RLN-Diff circuit defined in [RLN-v2](https://github.com/vacp2p/rfc-index/blob/dabc31786b4a4ca704ebcd1105239faff7ac2b47/vac/raw/rln-v2.md).
+The mapping is Sybil-resistant, requires no circuit changes, and enforces rate differentiation entirely through the membership registry.
+
+## 1. Introduction
+
+The [RLN Per-Hop DoS Protection](./mix-dos-protection-rln.md) assigns a flat rate limit to all mix nodes that meet a minimum stake requirement. To accommodate high-capacity relay nodes, this limit must be set high &mdash; making the same rate headroom available to any minimally-staked attacker. This structural limitation is referred to as the rate amplification gap, defined in [Section 3.1](#31-rate-amplification-gap).
```
jm-clius: Minor suggestion, but [semantic breaks](https://sembr.org/) throughout helps reviewing specific lines/thoughts. An example below:

```suggestion
The [RLN Per-Hop DoS Protection](./mix-dos-protection-rln.md) assigns a flat rate limit to all mix nodes that meet a minimum stake requirement.
To accommodate high-capacity relay nodes, this limit must be set high &mdash;
making the same rate headroom available to any minimally-staked attacker.
This structural limitation is referred to as the rate amplification gap, defined in [Section 3.1](#31-rate-amplification-gap).
```

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,243 @@
...
+
+`S_unit`, `R_base`, `T`, `R_min`, `R_max`, and `f` MUST be published in a deployment configuration accessible to all participants before the network accepts registrations.
+The membership registry MUST reject registrations inconsistent with the published parameters.
+Verifiers trust the membership registry to enforce the correct value of these parameters at registration time.
+
+## 5. Security and Privacy Considerations
+
+### 5.1 Sybil-Resistance
+
+The stake-to-rate mapping is Sybil-resistant: each rate increment of `T` costs `T × S_unit` of stake, regardless of how that stake is distributed across registrations.
```
jm-clius: Not a game changer, but worth considering: the punitive costs of abusing multiple registrations vs single high-rate registrations is potentially lower - for a total stake S over N registrations you will only get slashed S/N for each rate-limit violation, whereas for a single registration your entire S will get slashed.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,243 @@
...
+
+- **Dynamic stake top-up**: A mechanism for incrementally increasing stake without full deregistration would improve operational ergonomics.
+
+- **Rate-weighted path selection**: The [Mix Protocol](./mix.md) currently specifies uniform random path selection. Under stake-weighted rate limits, this distributes forwarding load equally regardless of node capacity, causing low-rate nodes to hit their limits and drop forwarding traffic. Adapting the [Mix Protocol](./mix.md) to weight path selection proportional to registered rates &mdash; similar to Tor &mdash; would address this.
+
+- **Cover traffic under non-uniform rates**: The [Mix Cover Traffic](./mix-cover-traffic.md) specification derives per-node cover emission bounds assuming uniform rate limits. Under stake-weighted rates, the forwarding load depends on the network's stake distribution, and the cover budget derivation does not directly apply.
```
jm-clius: My main concern/question with this approach lies in these two paragraphs. In fact, everything else neatly describes a mechanism for rate-limit differentiation, which makes sense to me :).
In short - my concern with a weighted path selection is that (even if we include reputation metrics), most well-funded adversaries will be able to exploit this significantly (and end up behaving better than other nodes) to have a greater chance of being inserted into more mix paths. I was convinced by recent discussions that purely random selection remains the best strategy against such surveillance attacks. To me, RLN rate differentiation doesn't imply a different design for network capacity - the mix network is designed around R_base, which is a function of what the total rate-limit should be given our network traffic expectations for legitimate use cases. Cover traffic should also be designed around "expected network traffic patterns" and is therefore a function of R_base.
Since we know that attackers could originate up to R_base of traffic (per membership) to target intermediate hops, we want to increase the cost of trivially "taking down" intermediate nodes by exploiting the fact that everyone starts dropping traffic if R_base is exceeded on their input. The way to mitigate this trivial attack is to allow some "deserving" nodes to be able to remain perfectly operational if input > R_base by increasing their rate-limit allocation, thus reducing the utility of the attack without changing our assumptions re network traffic and R_base. Admittedly, simply allowing higher rate-limits for higher stake is not enough and we'll need a reputation mechanism where very well-performing nodes can potentially reach rate-limits > R_max. I still wouldn't increase the probability of these nodes being included in path selection, though, as this increases the probability of a passive surveillance attack.

jm-clius (commented):
Thanks! On the whole, I think I agree with the spec. My only concern is about the implications discussed in the "Future Work" section. See my comments below.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
chaitanyaprem: Ok, i see the value in having uniform random path selection and keeping the default network rate limit at R_base. 
And agree that in future work, we should link reputation also as a means to increase rate limit and not just stake.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -0,0 +1,352 @@
...
+[RLN Per-Hop DoS Protection](./mix-dos-protection-rln.md) enforces a flat rate limit `R_base` per node per epoch on outgoing packets.
+Two properties combine to make this exploitable:
+
+- `R_base` must be set high enough to accommodate the forwarding load of high-capacity relay nodes.
+  As a result, all nodes receive the same high `R_base` regardless of stake.
+- The [Mix Protocol](./mix.md)'s unlinkability guarantees make forwarded and originated packets cryptographically indistinguishable by design,
+  so any node can use its `R_base` allowance entirely for origination.
+
+Together, these give a minimally-staked attacker the same effective origination budget
+as the forwarding budget intended for a high-capacity honest relay.
```
chaitanyaprem, line 103: I am wondering if it is a good idea to indicate what kind of attacks an attacker can perform due to this. That way a reader would understand and analyze how the spec is helping avoid such attacks and till what level. 
As of now it just reads that the only attack possible is dropping packets and use full budget for origination only which anyways doesnt get addressed via stake-weighted RLN.

The attack e.g like Hanno mentioned of trying to take down further intermediary nodes in a path becomes easier and hence whole network, whereas when using stake-weighted RLN it is harder to do as there are some nodes in the network with higher rate limit.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: True, per-violation cost drops from `S` to `S/N` when split across `N` registrations. But each split registration also has rate `R/N` (`N` times smaller than a single registration with stake `S`), so the cost per unit of rate lost is the same (`S/R`) in both cases. Splitting offers no advantage in slashing cost per rate-unit, except at the `R_max` ceiling already discussed in §5.1.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: > In short - my concern with a weighted path selection is that (even if we include reputation metrics), most well-funded adversaries will be able to exploit this significantly (and end up behaving better than other nodes) to have a greater chance of being inserted into more mix paths. I was convinced by recent discussions that purely random selection remains the best strategy against such surveillance attacks. 

Fully agree with the surveillance concern. A well-funded adversary could stake more to gain disproportionate path inclusion. Uniform random selection is a good defense against that.

One tension worth raising on the DoS side: path selection (uniform or weighted) only constrains honest senders. An attacker can stake the minimum and target low-stake intermediaries regardless of the path selection rule. The difference is in honest-path impact: weighted selection biases toward high-rate nodes, so only few honest paths are affected when low-stake intermediaries are flooded. Under uniform selection, honest paths uniformly include low-stake nodes, so flooding them degrades a larger share of honest traffic.

On the cover-traffic / `R_base` point: agreed that `R_base` anchors the design. To make this hold cleanly under stake-weighting, R_min is now derived as `⌈R_base/T⌉ × T`, so every node has `user_message_limit ≥ R_base` by construction. The cover-traffic derivation can use `R_base` as the common rate parameter across all nodes. Open question (flagged in §7): whether stake-weighted capacity in `(R_base, R_max]` can be utilized for emission without breaking uniformity — resolving this requires either accepting unused capacity above `R_base` or amending the cover-traffic derivation to accommodate budgets at least up to a fraction above `R_base`. @chaitanyaprem

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
AkshayaMani: Good point. I think the current §3.1 captures the attack capability abstractly: a minimally-staked attacker can use the entire `R_base` (set high for honest forwarding) for origination. Concretely, this manifests as intermediary flooding — pre-RLN this was trivial (unbounded); per-hop RLN bounds it to `R_base` per membership, but `R_base` is set high enough that this remains a useful attack budget (the rate amplification gap). Stake-weighted RLN raises the cost of higher-rate attacks (§5.1) but does not further reduce the per-membership `R_base` available for baseline-rate flooding (§5.2). As discussed in the path-selection thread, the "high-rate nodes absorb floods" mitigation only materializes under rate-weighted path selection, which we've rejected for surveillance reasons. Open to suggestions.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
chaitanyaprem: > Open question (flagged in §7): whether stake-weighted capacity in `(R_base, R_max]` can be utilized for emission without breaking uniformity — resolving this requires either accepting unused capacity above `R_base` or amending the cover-traffic derivation to accommodate budgets at least up to a fraction above `R_base`

Since we are sticking to uniform random path selection, it would make sense for cover-traffic to stick to R_base as max and normally work with fraction or `R_base` itself. Increasing cover traffic beyond `R_base` would  increase the probability of real packets getting dropped along the paths(just because a node in the path has `R_base` rate limit and it got selected by many other nodes that are generating cover traffic at a higher rate than `R_base`).  If all nodes generate at a fraction of `R_base` this probability reduces.  Also since this is just an input to cover traffic, we can keep it simple to begin with a use `R_base` itself.

AkshayaMani:
Agreed, `(R_base, R_max]` stays non-cover headroom, and #341 formalizes it.
Also, moved reputation-based rate differentiation to future work per the discussion above, and `R_max` is now published directly.

@jm-clius, planning to merge, been open a while and needs one approval to land.

chaitanyaprem (approved):
LGTM, Thanks!

PR #82, opened 2024-07-08
Add RLNv2 to TheWakuNetwork

alrevuelta:
* Add RLNv2 to TheWakuNetwork
* Set parameters as listed [here](https://github.com/waku-org/nwaku/blob/master/waku/factory/networks_config.nim#L31)

Issue #66, opened 2024-06-14
New RFC: 40/WAKU2-CAPABILITY-PUSH


jimstir:
Reference issue: https://github.com/vacp2p/rfc/issues/521
Author: kaiserd

This issue is part of the Waku ambient discovery roadmap (new roadmap issue to be opened in the research repo; will edit/update this description once the roadmap issue is open)
and tracks the editing of a new RFC specifying a Waku capability push protocol.
(also see #520)

## Background 

In its first version, it will simply be [libp2p identify/push](https://docs.libp2p.io/concepts/protocols/#identify)
with restrictions on specific fields.

Reasons for having a separate Waku spec instead of simply using libp2p identify/push may be found in #520.

Reasons for having a separte RFC for capability-push and not make it part of #520, even though these protocols are very similar with respect to the transmitted messages:

* these protocols are semantically different
* these protocols might develop in different directions
* imo, each RFC should only specify a single protocol ID.

(Sill, I could merge these two into one, if desired.)

PR #329, opened 2026-05-04
Add LIONESS specification for Mix payload encryption

mghazwi:
This PR includes the LIONESS wide-block encryption specification to be used by the Mix protocol for payload (delta) encryption. This replaces the use of AES-CTR and addresses the integrity problem raised in issue #306 

**LIONESS specification**
The LIONESS spec includes the following:
- the LIONESS construction,
- payload plaintext format,
- payload and LIONESS round keys derivation,
- payload construction,
- SURB reply payload construction,
- per-hop payload processing,
- payload integrity verification.

**Mix spec key changes:**
- replace AES-CTR with LIONESS for payload encryption, and update to the formulas to use LIONESS.
- update the per-hop packet processing so each mix node decrypts using LIONESS
- update exit processing so the final hop verifies the zero-prefix 
- updates SURB reply processing to account for the fact that LIONESS encryption and decryption are not the same operation as in AES-CTR. This requires the SURB creator to reverse the transformations done on the reply path.


docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+- the required payload format and construction.
+- the hop payload processing and integrity verification.
+
+## 1. Introduction
+
+The libp2p Mix Protocol uses a Sphinx packet format with four fields ($\alpha,\ \beta,\ \gamma,\ \delta$). The header fields ($\alpha,\ \beta,\ \gamma$) provide per-hop layered-encrypted routing information which includes header integrity. The payload field $\delta$ carries only the application message in layered-encrypted form.
+
+In the current Mix Protocol, AES-CTR is used to encrypt the routing header $\beta$. This is sufficient for $\beta$ because header integrity is separately protected by the per-hop MAC field $\gamma$. However, the integrity of $\delta$ is not covered by $\gamma$, since the Sphinx design intentionally separates header integrity from payload integrity.
+
+This separation is necessary because a valid Sphinx header may be paired with different payload content, and because SURB replies require the sender of the SURB to construct the return header before the reply payload is known. As a result, payload integrity MUST be provided independently of header integrity.
```
bkomuves: I think SURBs is the primary motivation here; otherwise in theory you could always construct a header after the payload is availabler. I would rephrase accordingly.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+
+A malleable encryption scheme such as AES-CTR does not satisfy this requirement. Bit modifications to the ciphertext result in modifications to the decrypted plaintext. This violates the integrity of the Sphinx payload. Therefore, the payload-encryption scheme for libp2p mix MUST satisfy the following:
+1. It MUST preserve the fixed payload size $|\delta|$.
+2. It MUST support layered encryption and per-hop layer removal.
+3. It MUST be compatible with SURBs and therefore MUST NOT require payload-dependent header authentication.
+4. It MUST allow the final hop to detect payload tampering.
+5. It MUST avoid adding external authentication material/tags that change the packet or payload size.
+
+To achieve this, this specification introduces LIONESS, a wide-block cipher built from a stream cipher and a keyed hash function. LIONESS acts as a pseudo-random permutation (PRP) over the entire payload block, allowing us to add an integrity prefix (e.g. leading zeros) into the plaintext and verify it after decryption.
```
bkomuves: This sounds like this is something new from us. I would refer the Lioness paper here, and make it clear that we just document an existing construction, and specify some concrete choices

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+
+## 2. Terminology
+The following terms are used throughout this specification:
+
+- **Mix Protocol**: The libp2p-mix protocol defined in [the Mix specification](./mix.md).
+- **Sphinx packet**: The packet format used by the Mix Protocol, consisting of ($\alpha,\ \beta,\ \gamma,\ \delta$).
+- **Payload**: The fixed-size encrypted field $\delta$ in a Sphinx packet.
+- **Payload integrity prefix**: A fixed all-zero prefix added to the plaintext payload before layered encryption and checked after final decryption.
+- **Wide-block cipher**: A cipher that operates on the whole payload block as one unit rather than as independent small blocks.
```
bkomuves: Technically a "wide block cipher" only means a large block size (compared to for example the fixed 128 bit blocks of AES). 

A wide block cipher can be still used as a block cipher, but here we don't want do that. Always applying it on a single block makes it a PRP.

Furthermore, I guess there could be wide block ciphers which do not allow arbitrary block size (eg. only multiple of 16 bytes). While it's not that important here, as we could just define the payload size to allow for such constraints, maybe it's worth mentioning that Lioness allows arbitrary block size (above a lower lound)

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+- **Payload encryption key ($\mathbf{\delta_{key}}$)**: A key derived from the per-hop shared secret. It is used to derive the four LIONESS internal round keys.
```
bkomuves: It's only derived when mix nodes process it. In SURBs, there is also a random payload key too, as part of the SURB.

(a SURB is a triple `(A,H,K)` where H is a header, A is the address of the entry node, and K is a random payload key. This makes reply message private and tamper-proof even when the sender is not a mix node; this is in contrast with forward messages where the exit node must be trusted)

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+### 3.1 Stream Cipher
+
+The stream cipher $\mathsf{S}$ used in LIONESS. We denote this as:
+$$
+\mathsf{S}(k, \mathsf{iv}, n) \to \mathsf{ks}
+$$
+
+where:
+- $k$ is a 32 byte key
+- $\mathsf{iv}$ is IV/nonce with size that depends on the chosen stream cipher.
```
bkomuves: I would bundle the IV and the key into a single "big key". This makes presentation simpler and less confusing (see below), and more uniform for stream ciphers without an IV. 

As all keys are essentially random and unique, the IV is not really relevant.

Furthermore, we are not dealing with arbitrary stream cipher, only XOR with a pseudo-random stream generator (PRG). So I would phrase instead that S takes a key and produces a infinite sequence of pseudo-random bytes. Then you don't have to deal with the length either.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+where:
+- $k$ is a 32-byte key
+- $m$ is an arbitrary-length input message.
+- $h$ is the 32-byte digest output
+
+### 3.3 Key Derivation Function (KDF)
+The key derivation function $\mathsf{KDF}$ is used to derive the internal LIONESS round keys from:
+- $\mathsf{dom}$: a domain-separation string
+- $\mathsf{seed}$: a 32-byte seed 
+- $\mathsf{len}$: a required output length  in bytes
```
bkomuves: I have a slight preference to ask for only (at most) 32 bytes derived key. The reason for this that you have a larger selection of KDFs (eg. a simple SHA256 would could also work, which is also simpler than a XOF).

As we only need 4 32 byte keys, you can derive them with 4 different domain-separator.

I would like to bound the length domain-separator string, eg. to 16 or 20 characters (bytes). So that a SHA256 key derivation calls the underlying SHA256 compression primitive only once.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+\mathsf{Lioness.Dec}(k,y) \to x 
+\end{array}
+$$
+where:
+- $k$ is the seed/master key from which the internal round keys are derived, with size $|k| = 32$ bytes
+- $x$ is the plaintext message with size $|x| \ge 2 |k|$ bytes.
+- $y$ is the corresponding ciphertext with size $|y| = |x|$
+
+### 4.2 Block Structure
+LIONESS is a wide-block cipher, meaning that it can take a large plaintext message and process it as a single large block by applying a small Feistel network over it. Instead of splitting the plaintexts into small blocks and processing these, LIONESS splits the input message block $b$ into two chunks:
```
bkomuves: Lioness in fact allows for splitting the plaintext to blocks, you could use it with any stream cipher mode of operation same as AES. 

It's just we don't want to use that way, but instead want to use it as a PRP

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+
+- $K_1$ and $K_3$ are the keys used for the stream cipher.
+- $K_2$ and $K_4$ are the keys used for the keyed hash function.
+
+All internal round keys are $|k| = 32$ bytes in size. Therefore, the required output key material from the $\mathsf{KDF}$ is 128 bytes. 
+
+We expect the seed/master key ($\delta_{key}$) to be of size 32 bytes. This $\delta_{key}$ is derived from the shared key $s_i$ as defined in the [mix specification](./mix.md). Therefore, we use this $\delta_{key}$ to derive all the LIONESS round keys to encrypt the payload. For this we use the KDF as defined in section 3.3:
+
+$$
+(K_1 \parallel K_2 \parallel K_3 \parallel K_4) = \mathsf{KDF}(\texttt{"lioness-payload-key"}, \delta_{key}, 128)
```
bkomuves: See above about deriving the 4 keys with 4 KDF calls instead (both is OK but that would allows for simpler KDF versions)

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+- $S$ is the stream cipher as defined in section 3.1
+- $H$ is the keyed hash function as defined in section 3.2 
+- $\mathsf{iv}$ is the initialization vector or nonce for the stream cipher. The choice of $\mathsf{iv}$  and its size depend on the chosen stream cipher and therefore depend on how LIONESS is instantiated.
+
+LIONESS encryption proceeds with applying a small Feistel network of four rounds:
+
+$$
+\begin{aligned}
+B &= L_0 \parallel R_0 \\
+R_1 &= R_0 \oplus S(K_1 \oplus L_0, \mathsf{iv}, |R_0|) \\
```
bkomuves: this is confusing as IV should also depend on K_1. See above just using a single (32 byte) key for S. 

Then the AES instance can be specified to split that into a 16 byte AES key and 16 byte IV/nonce

Furthermore, |R_0| = |R_1|, which again just make noise. I would just omit the lengths here, it's clear from the context

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+The resulting $\delta$ is placed into the final Sphinx packet.
+
+#### Reply payload (SURB payload)
+For a SURB reply, the reply sender does not know the return-path shared secrets $s_0, \ldots, s_{L-1}$.
+Therefore, the reply sender MUST perform the following steps:
+- constructs the payload plaintext:
+    $$
+    B = 0_\kappa \parallel m
+    $$
+- Then it derives the reply payload encryption key:
```
bkomuves: I don't think this initial reply encryption key is derived; on the original Sphinx paper it's a randomly generated key included in the SURB itself (so generated by the original sender).

(As an optimization, I will propose to derive it from the SURB header itself (which is essentially random anyway), so the SURB is 16-32 byte smaller. But that's a separate proposal, part of the larger compressed address proposal)

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+
+- [Rust reference implementation of LIONESS](https://github.com/mghazwi/lioness_blockcipher)
+- [Haskell reference implementation](https://github.com/logos-storage/transport-over-mix).
+
+These reference implementations are generic and can support any compatible $\mathsf{S}$, $\mathsf{H}_k$, and $\mathsf{KDF}$. 
+
+The current Mix protocol already uses primitives that can be used for LIONESS. A recommended instantiation is:
+- AES-CTR as the stream cipher $\mathsf{S}$,
+- HMAC-SHA-256 as the keyed hash $\mathsf{H}_k$,
+- HKDF-SHA256 as the key derivation function $\mathsf{KDF}$.
```
bkomuves: if we only require up to 32 bytes of derived keys (see above), then a simpler choice for KDF would be HMAC or even just `SHA256( domain || seed )`. The latter can be also used as a keyed hash.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,457 @@
...
+
+## 8. Reference Implementations
+
+- [Rust reference implementation of LIONESS](https://github.com/mghazwi/lioness_blockcipher)
+- [Haskell reference implementation](https://github.com/logos-storage/transport-over-mix).
+
+These reference implementations are generic and can support any compatible $\mathsf{S}$, $\mathsf{H}_k$, and $\mathsf{KDF}$. 
+
+The current Mix protocol already uses primitives that can be used for LIONESS. A recommended instantiation is:
+- AES-CTR as the stream cipher $\mathsf{S}$,
```
bkomuves: here maybe specify the splitting of the 32 byte stream key into 16 byte AES key (for AES128) and 16 byte IV.

(If larger AES key size is used (AES192 or AES256) then another key derivation would be required, but I think those choices are pointless here)

docs/ift-ts/raw/mix.md
```diff
@@ -635,9 +641,6 @@ The recommended total packet size is $4608$ bytes, chosen to:
 - Allow inclusion of additional data such as SURBs without requiring fragmentation,
 - Maintain reasonable per-hop processing and bandwidth overhead.
 
-Note:
-When [DoS protection](#66-dos-protection) is enabled, the maximum packet size MUST be increased to accommodate the proof size of the chosen DoS protection mechanism.
```
AkshayaMani, line 651: The deleted note was correctly scoped and doesn't belong in the _message size_ constraints below. A DoS-protection proof $σ$ is appended to the Sphinx packet (header + $δ$ + $σ$), and is not embedded inside $δ$. Therefore, DoS-protection proofs expand the wire-level _packet size_, not the maximum application _message size_ that fits inside $δ$. 

So this _packet size_ note should be restored.

docs/ift-ts/raw/mix.md
```diff
@@ -647,8 +650,10 @@ Payload &= 4608 - 624 \\
 \end{aligned}
 `$
 
-Implementations MUST account for payload extensions, such as SURBs, when determining the maximum message size that can be encapsulated in a single Sphinx packet.
-Details on SURBs are defined in [Section 8.7.1](#871-surb-component-sizes).
+Implementations MUST account for the following when determining the maximum message size that can be encapsulated in a single Sphinx packet: 
+- the zero-prefix authentication added to the payload as defined in the [LIONESS specification](./mix-lioness.md)
+- payload extensions, such as SURBs (see [Section 8.7.1](#871-surb-component-sizes)).
+- the proofs for [DoS protection](#66-dos-protection) in case it is enabled.
```
AkshayaMani: A couple of things to be considered:
- The bullet on proofs for [DoS protection](#66-dos-protection) doesn't belong here (see comment above).
- It might be cleaner to also list _Payload integrity prefix_ in `mix.md`'s Terminology section, so both specs reference the same definition.

The current sentence could then become:

> Implementations MUST account for payload integrity prefix size (defined in the [LIONESS specification](./mix-lioness.md)) and payload extensions, such as SURBs, when determining the maximum message size that can be encapsulated in a single Sphinx packet.

docs/ift-ts/raw/mix.md
```diff
@@ -741,7 +746,7 @@ The construction MUST proceed as follows:
...
+   - Pad the result to the maximum application message length of $3968$ bytes using a deterministic padding scheme. This value is derived from the fixed payload size in [Section 8.3.2](#832-payload-size) ($3984$ bytes) minus the payload zero-prefix with size equals to the security parameter $κ = 16$ bytes defined in [Section 8.2](#82-cryptographic-primitives). The chosen scheme MUST yield a fixed-size padded output and MUST be consistent across all mix nodes to ensure correct interpretation during unpadding. For example, schemes that explicitly encode the padding length and prepend zero-valued padding bytes MAY be used.
```
AkshayaMani: A few things to be considered:
- It might be better to use _payload integrity prefix_ instead of _payload zero-prefix_ for terminology consistency.
- The prefix size being $κ$ is a design choice. Better to state that explicitly.
- Might be clearer to clarify that the last line is about the _padding_ scheme.

Suggested:

> ...minus the payload integrity prefix size, which is set to $16$ bytes to match the security parameter $κ$ defined in [Section 8.2](#82-cryptographic-primitives). The chosen padding scheme...

docs/ift-ts/raw/mix.md
```diff
@@ -878,29 +883,26 @@ The construction MUST proceed as follows:
 
    d. **Encrypt Payload**
    The encrypted payload $δ$ contains the message $m$ defined in step 1, prepended with a $κ$-byte string of zeros.
```
AkshayaMani: Suggested:
>    The encrypted payload $δ$ contains the message $m$ defined in step 1, prepended with the payload integrity prefix of size $κ$.

docs/ift-ts/raw/mix.md
```diff
@@ -878,29 +883,26 @@ The construction MUST proceed as follows:
...
-   It is encrypted in layers such that each hop in the mix path removes exactly one layer using the per-hop session key.
+   It is encrypted in layers using the LIONESS wide-block encryption as defined in the [mix-lioness specification](./mix-lioness.md). Each hop in the mix path removes exactly one layer using the per-hop session key as the seed to derive all LIONESS internal round keys.
```
AkshayaMani: Suggested:

> It is encrypted in layers using the LIONESS wide-block encryption (defined in the [LIONESS specification](./mix-lioness.md)) such that each hop in the mix path removes exactly one layer using the per-hop session key. 

- the detail on how internal round keys are derived belongs in the [LIONESS specification](./mix-lioness.md).

docs/ift-ts/raw/mix.md
```diff
@@ -878,29 +883,26 @@ The construction MUST proceed as follows:
...
-     \mathrm{KDF}(\text{"δ\_aes\_key"} \mid s_i)\\
-     δ_{\mathrm{iv}_i} =
-     \mathrm{KDF}(\text{"δ\_iv"} \mid s_i)
+     δ_{\mathrm{key}_i} =
+     \mathrm{KDF}(\text{"delta\_key"} \mid s_i)\\
      \end{array}
      `$
 
-   - Using the derived keys, compute the encrypted payload $δ_i$:
+   - Apply the LIONESS payload encryption step defined by that specification to compute the encrypted payload $δ_i$:
```
AkshayaMani: Suggested (optional):

> Using the derived keys, compute the encrypted payload $δ_i$ using LIONESS encryption (see [LIONESS specification](./mix-lioness.md)):

docs/ift-ts/raw/mix.md
```diff
@@ -909,8 +911,8 @@ The construction MUST proceed as follows:
 
        $`
        \begin{array}{l}
-       δ_i = \mathrm{AES\text{-}CTR}\bigl(δ_{\mathrm{aes\_key}_i},
-       δ_{\mathrm{iv}_i}, δ_{i+1} \bigr)
+       δ_i = \mathrm{Lioness.Enc}\bigl(δ_{\mathrm{key}_i},
```
AkshayaMani: Optional polish: Use `LIONESS.Enc` / `LIONESS.Dec` (all caps) here and throughout (across both files), to match
- the all-caps usage of `LIONESS` elsewhere in this spec body, and
- the existing `AES-CTR` convention in `mix.md`.

docs/ift-ts/raw/mix-lioness.md
mghazwi: I agree, I was trying to simplify. at some point i considered writing a definition of PRP, but didn't. I'll probably add that to clarify

docs/ift-ts/raw/mix-lioness.md
mghazwi: Yes, the random payload key which is called reply key in the mix spec is included in the SURB, but in the Mix spec and implementation it is 16 bytes and used to derived the payload AES key and IV in the same way as shared secret. This not the same as the paper.

docs/ift-ts/raw/mix-lioness.md
mghazwi: yeah that makes sense, I was trying to make it more general for any KDF.
yes bounding the domain separator string to 16 bytes should be enough.

docs/ift-ts/raw/mix-lioness.md
mghazwi: yeah i had it without length before, but thought that would avoid confusion about lengths, but yeah it is clear from context.

docs/ift-ts/raw/mix-lioness.md
mghazwi: Yes in the paper it is not derived, but in the current Mix spec and implementation it is derived so this should be consistent with the mix spec. 
The reply encryption key (reply/surb key) is 16 bytes, and in spec/code they derive the AES key and IV from that.

docs/ift-ts/raw/mix-lioness.md
mghazwi: yes will do.

docs/ift-ts/raw/mix-lioness.md
mghazwi: i was thinking that, but it seemed simpler to just abstract KDF. Internally the KDF can call some internal function 4 times to generate the required key material e.g. 4 SHA256 calls

docs/ift-ts/raw/mix.md
```diff
@@ -1316,9 +1315,21 @@ Once the destination responds with a reply message, the Exit Layer MUST perform
 
 2. **Encrypt Reply Payload**
 
-   Derive the AES key, IV, and compute the encrypted payload $δ$ following the same procedure in [Section 8.5.2](#852-construction-steps) Step 3.d for $i = L-1$, substituting $\tilde{k}$ for $s_{L-1}$.
-
-   Note: Unlike the forward path ([Section 8.5.2](#852-construction-steps) Step 3.d), the reply payload is encrypted only once&mdash;using $\tilde{k}$ alone. Each mix node on the return path will apply one additional layer during Sphinx processing ([Section 8.6.1](#861-shared-preprocessing) Step 5), resulting in $L + 1$ layers total that the initiating node must remove during reply recovery (see [Section 8.7.5](#875-reply-recovery)).
+   Unlike the forward path ([Section 8.5.2](#852-construction-steps) Step 3.d), the reply payload is encrypted only once using $\tilde{k}$ alone. Therefore, the mix node MUST perform the following steps:
```
AkshayaMani: Pattern across §8.7.3 Step 2 and §8.7.5 Step 2: They duplicate crypto procedures already defined elsewhere with minor changes. Reference-style is cleaner: single source of truth, easier to maintain, and avoids the small bugs that have crept in both sections. 
Consider reverting them, with just the AES → LIONESS terminology updates inline.

docs/ift-ts/raw/mix.md
```diff
@@ -1440,6 +1446,7 @@ The core Mix Protocol&mdash;comprising anonymous routing through a sequence of m
 - **Metadata protection**: All messages are fixed in size and indistinguishable on the wire. Sphinx packets reveal only the immediate next hop and delay to each mix node. No intermediate node learns its position in the path or the total path length.
 - **Traffic analysis resistance**: Continuous-time mixing with randomized per-hop delays reduces the risk of timing correlation and input-output linkage.
 - **Per-hop confidentiality and integrity**: Each hop decrypts only its assigned layer of the Sphinx packet and verifies header integrity via a per-hop MAC.
```
AkshayaMani: Suggested:
> - **Per-hop header confidentiality and integrity**: Each hop decrypts only its assigned layer of the Sphinx header and verifies header integrity via a per-hop MAC.
> - **Per-hop payload confidentiality**: Each hop removes only one layer of encryption, so intermediate hops cannot recover the plaintext. 
> - **End-to-end payload integrity**: The exit node verifies the payload integrity prefix after fully decrypting the payload. For more details see the [LIONESS specification](./mix-lioness.md).

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+- $|m| = |B| - |z|$
+
+The size of the payload $|\delta|$ is specified in the Mix protocol, and if the application message $|p|$ is small payload padding is added as specified in the [Mix Protocol](./mix.md).
+
+### 5.2 Sphinx Payload Construction
+
+#### Forward Payload
+Once the plaintext is formatted as specified above, it needs to be encrypted in layers such that each hop in the mix path removes exactly one layer using the per-hop session key. This ensures that only the final hop (i.e., the exit node) can fully recover the plaintext message $m$, validate its integrity, and forward it to the destination. To compute the encrypted payload, perform the following steps for each hop $i = L-1$ down to $0$, recursively:
+
+- Derive the payload key $\delta_{\mathrm{key}_i} = \mathsf{KDF}(\texttt{"delta_key"}, s_i, 32)$ where $s_i$ is the per-hop shared secret for hop `i` as defined in the Mix protocol specification.
```
chaitanyaprem: noticing some parse errors here and some other places. Example noticed in vscode. Also noticed in github parsing.

`Derive the payload key ParseError: KaTeX parse error: Expected 'EOF', got '_' at position 54: …(\texttt{"delta_̲key"}, s_i, 32) where s_i is the per-hop shared secret for hop i as defined in the Mix protocol specification.`

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+## 3. Cryptographic Primitives
+
+This section defines the primitives used by this specification. In this specification, we will assume the following constants:
+- The security parameter $\kappa = 16$ bytes ($128$-bits)
+- The key size $k = 32$ bytes.
+
+### 3.1 Stream Cipher
+
+The stream cipher $\mathsf{S}$ used in LIONESS. We denote this as:
+$$
```
chaitanyaprem: This line is not rendering properly in github, but renders fine in vscode. Maybe there is some diff in parser used. Similarly lot of such lines are not rendered correctly. 

Not sure if they would be fine when it is rendered in lip.logos.co, but better to check and fix.

docs/ift-ts/raw/mix-lioness.md
bkomuves: i mean 16 bytes is most probably fine (just pad with zeros, it has enough entropy), but we are not doing AES for payload encryption (that's the whole point :) so i don't think it matters what was written there before...

docs/ift-ts/raw/mix-lioness.md
bkomuves: unfortunately github's markdown parser is like _extremely broken_ especially with latex. 

(but i haven't checked for actual errors here)

docs/ift-ts/raw/mix-lioness.md
bkomuves: lip.logos.co uses a different parser, it should be fine there.

(but maybe add an empty line before `$$`)

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+- the required payload format and construction.
+- the hop payload processing and integrity verification.
+
+## 1. Introduction
+
+The libp2p Mix Protocol uses a Sphinx packet format with four fields ($\alpha,\ \beta,\ \gamma,\ \delta$). The header fields ($\alpha,\ \beta,\ \gamma$) provide per-hop layered-encrypted routing information which includes header integrity. The payload field $\delta$ carries only the application message in layered-encrypted form.
+
+In the current Mix Protocol, AES-CTR is used to encrypt the routing header $\beta$. This is sufficient for $\beta$ because header integrity is separately protected by the per-hop MAC field $\gamma$. However, the integrity of $\delta$ is not covered by $\gamma$, since the Sphinx design intentionally separates header integrity from payload integrity.
+
+This separation is necessary SURB replies require the sender of the SURB to construct the return header before the reply payload is known. As a result, payload integrity MUST be provided independently of header integrity.
```
AkshayaMani: missing "because".

Suggested:

> This separation is necessary because SURB replies require the sender of the SURB to construct the return header before the reply payload is known.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+A malleable encryption scheme such as AES-CTR does not satisfy this requirement. Bit modifications to the ciphertext result in modifications to the decrypted plaintext. This violates the integrity of the Sphinx payload. Therefore, the payload-encryption scheme for libp2p mix MUST satisfy the following:
+1. It MUST preserve the fixed payload size $|\delta|$.
+2. It MUST support layered encryption and per-hop layer removal.
+3. It MUST be compatible with SURBs and therefore MUST NOT require payload-dependent header authentication.
+4. It MUST allow the final hop to detect payload tampering.
+5. It MUST avoid adding external authentication material/tags that change the packet or payload size.
+
+To achieve this, this specification uses LIONESS as described in [anderson et al](https://www.cl.cam.ac.uk/archive/rja14/Papers/bear-lion.pdf). LIONESS is a wide-block cipher built from a stream cipher and a keyed hash function. LIONESS acts as a pseudo-random permutation (PRP) over the entire payload block, allowing us to add an integrity prefix (e.g. leading zeros) into the plaintext and verify it after decryption.
+
+## 2. Terminology
```
AkshayaMani: A few thoughts on keeping the spec aligned with `mix.md` and avoiding terminology drift:
- Where a term or symbol is already defined in `mix.md`, suggest reusing the existing definition rather than introducing a parallel one here. Keeps the two specs consistent and reduces the load on implementors who have to read both.
- When this spec introduces a new term that's also used in `mix.md` (_e.g.,_ _Payload integrity prefix_), consider moving it to `mix.md`'s Terminology section (and updating `mix.md` to adopt the term) so both specs point to one definition.
- Otherwise, keep the Terminology section here scoped to truly local terms — and double-check there are no naming conflicts with `mix.md`.
- Use the canonical terms consistently throughout the body.

A few specific instances:
- KDF is defined as a fixed-length primitive in `mix.md` but as a variable-length primitive here. Consider either updating `mix.md`'s definition to a variable-length form or rename this to `LIONESS.KDF`.
- _Payload integrity prefix_ and _Payload encryption key_ are introduced here but also used in `mix.md` — could move both to `mix.md`'s Terminology.
- Both terms have several looser phrasings in the body — worth normalizing.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+
+- **Mix Protocol**: The libp2p-mix protocol defined in [the Mix specification](./mix.md).
+- **Sphinx packet**: The packet format used by the Mix Protocol, consisting of ($\alpha,\ \beta,\ \gamma,\ \delta$).
+- **Payload**: The fixed-size encrypted field $\delta$ in a Sphinx packet.
+- **Payload integrity prefix**: A fixed all-zero prefix added to the plaintext payload before layered encryption and checked after final decryption.
+- **Wide-block cipher**: A block cipher with a large block size compared to conventional fixed-size block ciphers such as AES. LIONESS is a wide-block cipher that supports variable-length input blocks above a lower bound.
+- **Payload encryption key ($\mathbf{\delta_{key}}$)**: A key derived from the per-hop shared secret. It is used to derive the four LIONESS internal round keys.
+- **Round keys**: The four keys $(K_1, K_2, K_3, K_4)$ used by the LIONESS Feistel network.
+
+## 3. Cryptographic Primitives
```
AkshayaMani: A couple of things worth pinning:
- Concrete primitives: `S` and `H_k` are both abstract — §8 lists recommended instantiations but doesn't pin them. Worth pinning these explicitly, so two implementations don't pick different primitives and silently fail to interoperate.
- Stream-cipher consumption rule: Depending on the cipher chosen, the 32-byte input `k` may need to be split into a `key` and `IV`. Worth capturing the split rule explicitly, including the byte order.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+LIONESS is a wide-block cipher, meaning that it can take a large plaintext message and process it as a single large block by applying a small Feistel network over it. Instead of splitting the plaintexts into small blocks and processing these, LIONESS splits the input message block $b$ into two chunks: 
+$$
+B = L \parallel R 
+$$
+where:
+- $B$ is the plaintext message block of any size $|B| \ge 2 |k|$
+- $L$ is the left chunk with size $|L| = |k|$ bytes
+- $R$ is the right chunk with size $|R| = |m| - |k|$ bytes. 
+
+LIONESS scheme requires the size of the message to be at least $|k| + \kappa$, i.e., $|B| \ge |k| + \kappa$ where $\kappa$ is the security parameter.  For simplicity, in this specification, we require the payload size to be strictly greater than 64 bytes (i.e., $\ge 2 |k|$). The Mix protocol payload size satisfies this requirement since the expected payload is much larger than 64 bytes.
```
AkshayaMani: A few follow-ups on this paragraph:
- "strictly greater than" contradicts $≥ 2|k|$ (greater than or equal to).
- "For simplicity, ..." — the simplification of $|k| + κ$ vs $2|k|$ is not clear. Maybe add a one-liner explanation.
- "The Mix protocol payload size satisfies this requirement…" — true with current design choices. Worth adding this it to `mix.md`, so the requirement is visible to any future change.

docs/ift-ts/raw/mix-lioness.md
```diff
@@ -0,0 +1,454 @@
...
+    $$
+3. parse the decrypted payload $\delta'$ as $B = z \parallel m$, where $|z| = \kappa$ bytes.
+4. verify that the first $\kappa$ bytes of $B$ are all zeros.
+5. discard the packet if this integrity check fails.
+6. otherwise remove the $\kappa$ bytes prefix and pass $m$ to the Mix Exit Layer.
+
+**Exit Processing - Reply packet**
+If the node is the exit, and the packet is a reply, it MUST:
+1. reverse the return-path transformations, i.e., since the hops apply LIONESS decryption, the exit must apply LIONESS encryption:
+    For each hop `$i = L-1$` down to `$0$`:
```
AkshayaMani: Formatting nits (Optional polish):
- `$i = L-1$` and `$0$` shouldn't have backticks — they might break math rendering.
- `$$...$$` display math doesn't always render reliably on GitHub. `mix.md` uses inline-math-with-backticks plus a `\begin{array}{l}` block for display math — consider switching for consistency and reliability.

AkshayaMani (commented):
This looks great! 🚀 Left a few comments mainly on terminology alignment with `mix.md`, pinning concrete primitives, and a few paragraph-level fixes.

docs/anoncomms/raw/mix-lioness.md
```diff
@@ -0,0 +1,548 @@
...
+
+A malleable encryption scheme such as AES-CTR does not satisfy this requirement. Bit modifications to the ciphertext result in modifications to the decrypted plaintext. This violates the integrity of the Sphinx payload. Therefore, the payload-encryption scheme MUST satisfy the following:
+1. It MUST preserve the fixed payload size $|\delta|$.
+2. It MUST support layered encryption and per-hop layer removal.
+3. It MUST be compatible with SURBs and therefore MUST NOT require payload-dependent header authentication.
+4. It MUST allow the final hop to detect payload tampering.
+5. It MUST avoid adding external authentication material/tags that change the packet or payload size.
+
+To achieve this, this specification uses LIONESS as described in [anderson et al](https://www.cl.cam.ac.uk/archive/rja14/Papers/bear-lion.pdf). LIONESS is a wide-block cipher built from a stream cipher and a keyed hash function. LIONESS acts as a pseudo-random permutation (PRP) over the entire payload block, allowing us to add an integrity prefix (e.g. leading zeros) into the plaintext and verify it after decryption.
```
AkshayaMani: One more terminology substitution:
> ...allowing us to add the payload integrity prefix into the plaintext and verify it after decryption.

docs/anoncomms/raw/mix-lioness.md
```diff
@@ -0,0 +1,548 @@
...
+- $\mathsf{seed}$: is an arbitrary-length seed at least $\kappa$ bytes in size.
+- $u$ is a $\mu$-byte output key.
+
+## 4. Concrete Primitive Instantiation
+
+This section defines the concrete primitive choices used by this specification.
+
+### 4.1 Stream Cipher Instantiation
+
+The LIONESS stream cipher $\mathsf{S}$ is instantiated using AES-CTR.
```
AkshayaMani, line 128: As @bkomuves mentioned, worth pinning the `AES-CTR` counter format explicitly (in `mix.md`). May be deferred to a later PR on `mix.md`.

docs/anoncomms/raw/mix-lioness.md
```diff
@@ -0,0 +1,548 @@
...
+
+$`
+\begin{array}{l}
+\mathsf{LIONESS.Enc}(k,x) \to y \\
+\mathsf{LIONESS.Dec}(k,y) \to x 
+\end{array}
+`$
+
+where:
+- $k$ is the arbitrary-length seed (master key) from which the internal round keys are derived, with size at least $\kappa$ bytes
```
AkshayaMani: seed (master key) → LIONESS seed 

for consistency.

docs/anoncomms/raw/mix-lioness.md
```diff
@@ -0,0 +1,548 @@
...
+
+Because the payload is encrypted as a single block, LIONESS acts as a pseudo-random permutation (PRP) over the entire payload, a modification to the ciphertext will, except with negligible probability, produce a decrypted plaintext whose first $\kappa = 16$ bytes are not all zero. Replacing LIONESS with a malleable stream construction invalidates this property.
+
+### 8.3 Primitive Choices
+The security of LIONESS depends on the security of the primitives used to instantiate it:
+- the stream cipher $\mathsf{S}$,
+- the keyed hash function $\mathsf{H}_k$,
+- the key derivation function $\mathsf{KDF}$.
+
+Implementations may use any compatible choices. For detailed analysis on the security of LIONESS, refer to the [paper](https://www.cl.cam.ac.uk/~rja14/Papers/bear-lion.pdf).
```
AkshayaMani: > Implementations may use any compatible choices.

§4 now specifies the concrete instantiation. Maybe drop it?

docs/anoncomms/raw/mix-lioness.md
```diff
@@ -0,0 +1,548 @@
...
+A malleable encryption scheme such as AES-CTR does not satisfy this requirement. Bit modifications to the ciphertext result in modifications to the decrypted plaintext. This violates the integrity of the Sphinx payload. Therefore, the payload-encryption scheme MUST satisfy the following:
+1. It MUST preserve the fixed payload size $|\delta|$.
+2. It MUST support layered encryption and per-hop layer removal.
+3. It MUST be compatible with SURBs and therefore MUST NOT require payload-dependent header authentication.
+4. It MUST allow the final hop to detect payload tampering.
+5. It MUST avoid adding external authentication material/tags that change the packet or payload size.
+
+To achieve this, this specification uses LIONESS as described in [anderson et al](https://www.cl.cam.ac.uk/archive/rja14/Papers/bear-lion.pdf). LIONESS is a wide-block cipher built from a stream cipher and a keyed hash function. LIONESS acts as a pseudo-random permutation (PRP) over the entire payload block, allowing us to add an integrity prefix (e.g. leading zeros) into the plaintext and verify it after decryption.
+
+## 2. Terminology
```
AkshayaMani, line 40: Consider removing terms now defined in `mix.md`'s Terminology: Sphinx payload ($δ$), Payload integrity prefix, Payload encryption key ($δ_{key}$).

docs/anoncomms/raw/mix.md
```diff
@@ -1383,70 +1402,39 @@ When the Exit Layer receives decrypted payload $δ'$ and the SURB identifier $\m
...
-     - If $i = 0$ (_i.e.,_ first layer):
-
-       $`
-       \begin{array}{l}
-       δ_i = \mathrm{AES\text{-}CTR}\bigl(δ_{\mathrm{aes\_key}_i},
-       δ_{\mathrm{iv}_i}, δ'
-       \bigr)
-       \end{array}
-       `$
+   The encrypted payload $δ'$ contains the of the padded reply message $m$ defined in [Section 8.7.3](#873-using-a-surb) Step 1, prepended with the payload integrity prefix of size $κ$-byte.
```
AkshayaMani: Typo:
> ...contains the padded...

docs/anoncomms/raw/mix.md
```diff
@@ -1383,70 +1402,39 @@ When the Exit Layer receives decrypted payload $δ'$ and the SURB identifier $\m
...
+   It is encrypted in a total of $L+1$ layers: 
+   A first layer of LIONESS encryption using the reply key $\tilde{k}$ as the seed, and
+   $L$ layers of LIONESS decryption using per-hop session keys $s_0, \ldots, s_{L-1}$.
+   To recover the padded reply message, we must first remove all the decryption layers with a LIONESS encryption, and then remove the initial reply encryption (using $\tilde{k}$) with a decryption. Therefore, the we must perform the following steps:
```
AkshayaMani: Typo:
>...Therefore, we must...

docs/anoncomms/raw/mix-lioness.md
```diff
@@ -0,0 +1,548 @@
...
+K_4 &= \mathsf{KDF}(\texttt{"lioness\_key4"}, k)
+\end{aligned}
+`$
+
+The LIONESS seed is arbitrary-length and MUST be at least $\kappa$ bytes. The derivation of the seed depends on whether the packet is a forward packet or a reply packet. Further details are specified in [Section 6](#6-payload-construction) and [section 7](#7-sphinx-payload-processing).
+
+### 5.4 Encryption
+
+Let:
+- plaintext message $B = L_0 \parallel R_0$
```
AkshayaMani, line 262: Clarification (minor):
- L262: Reader has to infer $|L_0|$ and $|R_0|$ (from §5.2).
- L264: Stream cipher $S$ is defined as producing arbitrary-length keystream (§3.1), but truncation to $|R|$ per round is implicit.

Maybe worth clarifying explicitly.

AkshayaMani (approved):
Thanks for the thorough revision! 🚀 A few follow-ups inline (mostly clarifications and a couple of typos).

docs/anoncomms/raw/mix-lioness.md
mghazwi: I will leave this for later PR as I think the current AES-CTR might change for hardware accelerated one, and I'm not sure if that changes the counter format as well or not.

PR #330, opened 2026-05-05
Mix hidden service specification

mghazwi:
This PR adds a draft specification for Hidden Services over Mix. 
The spec is still in early stage and needs more work. 

**Hidden Services over Mix**
The design is inspired by Tor onion services but adapted for Sphinx-style mixnets i.e. the mix protocol.
The spec includes the following:
- the hidden service workflow over Mix.
- establishing introduction points
- hidden service long-term identity keys `.mix` address
- descriptor contents, keys, and its periodic publication
- descriptor publication, retrieval, and validation using epoch-specific blinded keys
- bootstrapping a connection via intro point
-  appendix for Ed25519-style signing with key blinding

**TODO**
- Security considerations
- Restricted discovery mode
- DOS/spam prevention
- exact transport-layer interaction details
- experimentation in code to test if any flaws exist in the spec.


Cofson (commented):
Thank you for this comprehensive draft of the hidden service specification — it's a substantial piece of design work. I appreciate the explicit TODO markers throughout that signal where Security Considerations, restricted discovery, DOS/spam prevention, and replay scoping are still open. During my review, I had the following suggestions for the next iteration(s).

docs/anoncomms/raw/mix-hidden-service.md
```diff
@@ -0,0 +1,948 @@
...
+We can summarize the steps to encrypt the inner layer (i.e. the `EncryptedBlob`) as follows:
+
+```=
+secret_input = "inner layer" || hs_id || bk_i || revision_counter
+(enc_key, enc_iv, mac_key) = KDF(secret_input)
+salt = random()
+blob = streamcipher.enc(inner_layer, enc_key, enc_iv)
+mac = mac(salt || blob, mac_key)
+encrypted_blob = EncryptedBlob.new(salt, blob, mac)
+```
```
Cofson, line 628: **Suggestion (Section 8.4 — inner-layer KDF inputs):** `secret_input = "inner layer" || hs_id || bk_i || revision_counter` produces a deterministic `(enc_key, enc_iv)` pair for a given epoch + `revision_counter`. If a service ever re-encrypts different `inner_layer` contents under the same `revision_counter` (e.g., bug in the counter bump, restart with persisted state, or `revision_counter` wraparound), AES-128-CTR with the same `(key, iv)` leaks plaintext XOR catastrophically. The fresh `salt` is currently used only for the MAC and is not folded into the KDF. Consider one of: (a) include `salt` in `secret_input` so each encryption derives a fresh keystream; or (b) switch to an AEAD with a random nonce. Either change would also let you drop the separate `mac_key` / explicit MAC step.

docs/ift-ts/raw/mix-hidden-service.md
```diff
@@ -0,0 +1,948 @@
...
+
+### 9.2 Client behaviour
+To bootstrap a connection to a hidden service, the client first fetches and validates the hidden service descriptor as described in Section 8. After successful validation, the client obtains a list of introduction points, each containing:
+- intro point information `info`
+- authentication key `auth_key`
+- service key `service_key`
+
+The client then proceeds as follows:
+1. Select a random intro point from the list. Clients must select an intro point at random to distribute load and reduce linkability.
+2. Generate a fresh client encryption key `client_key`. The client must create a different `client_key` for every introduction attempt.
```
Cofson: **Suggestion (Section 9.2 step 2 — `client_key` randomness):** "Generate a fresh client encryption key `client_key`" — consider making the randomness requirement explicit, e.g., "MUST sample `client_key` uniformly at random from the DH keyspace for every introduction attempt." This matters because Section 9.5 derives `client_tag = H(client_key)` for the replay cache: if `client_key` were derived deterministically (e.g., from session state or a counter), tags would be predictable and an on-path adversary at the intro point could pre-fill the cache to deny service to a real client.

docs/anoncomms/raw/mix-hidden-service.md
```diff
@@ -0,0 +1,948 @@
...
+- Consume client-SURBs stored at the hidden service end (again this is dependent on how the transport layer manages SURBs)
+- Create duplicate application requests.
+
+In this protocol, replay resistance is achieved by the following:
+- *Binding to `auth_key`*: the MAC in `Introduce1` is computed over `auth_key || client_key || enc_payload`, therefore, an attacker cannot move a valid encrypted intro payload from one intro point to another or change `auth_key` without invalidating MAC. 
+- *Client tags*: The intro point and service must maintain replay cache/table of previously seen client tags for each service-intro session/state. These client tags are computed as `client_tag = H(client_key)`. When a request is received by either the intro point or service, they will check the existence of such tag in the local cache/table. If not present, record the tag in the cache/table and continue processing. If present, the request must be dropped with no further processing. The local cache for each service-intro session can be flushed with it expires or gets terminated (i.e. when `auth_key` is no longer assigned to a service/intro).
```
Cofson, line 827: **Suggestion (Section 9.5 — replay-cache retention horizon):** "The local cache for each service-intro session can be flushed with it expires or gets terminated (i.e. when `auth_key` is no longer assigned to a service/intro)" — consider specifying a minimum retention horizon. As written, retention is tied to session/`auth_key` lifetime, but an `auth_key` rotation that happens before the previous intro request fully expires from the network (e.g., delayed Sphinx delivery, retried sends) could allow late-replay against a successor `auth_key`. Tying retention explicitly to *at least* the intro-point validity window plus a delivery-delay margin would make the bound clearer for implementers.

docs/ift-ts/raw/mix-hidden-service.md
```diff
@@ -0,0 +1,948 @@
...
+*Note: Since $r$ is public in this protocol, compromise of $x'$ also compromises the long-term signing scalar $x$.*
+
+$$
+\begin{aligned}
+z \leftarrow \mathsf{Clamp}(r) \\
+x = z^{-1} x' \bmod \ell
+\end{aligned}
+$$
+
+*For this reason, the protocol uses blinded keys mainly to "certify" short-term descriptor signing keys, and keeps the long-term identity key and precomputed blinded signing material offline.*
```
Cofson: **Suggestion (Appendix — promote key-compromise note):** the property that public $r$ makes $x$ recoverable from $x$ (and therefore that compromising any per-epoch blinded signing scalar compromises the long-term identity) is a load-bearing architectural property — it drives the offline-storage requirement for `hs_id` and the precomputed `bk_i` material in Section 8.5 step 6. Currently this is only stated as an italic note in the appendix. Consider duplicating (or cross-referencing it explicitly) into Section 8.3 where blinded-key derivation is introduced, so a reader who skips the appendix still encounters the rationale for the offline-key handling.

docs/anoncomms/raw/mix-hidden-service.md
mghazwi: Thanks, I have address this in the last commit.

docs/ift-ts/raw/mix-hidden-service.md
mghazwi: modified the text slightly to highlight it.

docs/ift-ts/raw/mix-hidden-service.md
mghazwi: aside from the appendix, this is mentioned in section 8.4

PR #331, opened 2026-05-05
docs(messaging): moved segmentation spec

igor-sirotin:
Moved SEGMENTATION spec from https://github.com/logos-messaging/specs.
Original PR: https://github.com/logos-messaging/specs/pull/91

Changes included:
- frontmatter format
- convert links
- fix RFC2119 terminology styling (`MUST` instead of `**MUST**`)
- typos fixed

### DIFF
--- a/docs/messaging/raw/segmentation.md
+++ b/docs/messaging/raw/segmentation.md
@@ -1,0 +1,223 @@
+# MESSAGE-SEGMENTATION-AND-RECONSTRUCTION
+
+| Field | Value |
+| --- | --- |
+| Name | Message Segmentation and Reconstruction |
+| Version | 0.1 |
+| Status | raw |
+| Type | RFC |
+| Category | Standards Track |
+| Tags | segmentation |
+
+## Abstract
+
+This specification defines an application-layer protocol for **segmentation** and **reconstruction** of messages carried over a transport/delivery service with a message-size limitation, when the original payload exceeds said limitation.
+Applications partition the payload into multiple transport messages and reconstruct the original on receipt,
+even when segments arrive out of order or up to a **predefined percentage** of segments are lost.
+The protocol optionally uses **Reed–Solomon** erasure coding for fault tolerance.
+All messages are wrapped in a `SegmentMessageProto`, including those that fit in a single segment.
+
+## Motivation
+
+Many message transport and delivery protocols impose a maximum message size that restricts the size of application payloads.
+For example, Waku Relay typically propagates messages up to **150 KB** as per [64/WAKU2-NETWORK - Message](../core/64/network.md#message-size).
+To support larger application payloads, a segmentation layer is required.
+This specification enables larger messages by partitioning them into multiple envelopes and reconstructing them at the receiver.
+Erasure-coded parity segments provide resilience against partial loss or reordering.
+
+## Terminology
+
+- **original payload**: the full application payload before segmentation.
+- **data segment**: one of the partitioned chunks of the original message payload.
+- **parity segment**: an erasure-coded segment derived from the set of data segments.
+- **segment message**: a wire-message whose `payload` field carries a serialized `SegmentMessageProto`.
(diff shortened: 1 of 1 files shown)


docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,223 @@
...
+
+### Sending
+
+To transmit a payload, the sender:
+
+- MUST compute a 32-byte `entire_message_hash = Keccak256(original_payload)`.
+- MUST split the payload into one or more data segments,
+  each of size up to `segmentSize` bytes.
+  A payload of size ≤ `segmentSize` produces a single data segment (`data_segment_count == 1`).
+- MUST pad the last segment to `segmentSize` for Reed-Solomon erasure coding (only if Reed-Solomon coding is enabled).
```
Cofson, line 102: Padding scheme is unspecified. The sender MUST pad the last segment to `segmentSize` when Reed-Solomon is enabled, but the spec doesn't define how (zero-pad, PKCS#7, length-prefix?), and the Receiving section doesn't say how to strip padding before computing the hash. The `Keccak256(reconstructed_payload)` check at line 124 will fail unless both sides agree on the exact scheme. Two reasonable fixes: (a) pin the scheme explicitly (e.g. 'pad with zero bytes; the receiver strips trailing zeros before hashing' — but this breaks for payloads that legitimately end in zeros), or (b) add an `original_payload_length` field to `SegmentMessageProto` so the receiver can truncate to the exact byte count before hashing.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,223 @@
...
+
+A message is either a **data segment** (when `is_parity == false`) or a **parity segment** (when `is_parity == true`).
+
+### Validation
+
+Receivers MUST enforce:
+
+- `entire_message_hash.length == 32`
+- `data_segment_count >= 1`
+- `data_segment_count + parity_segment_count < maxTotalSegments`
```
Cofson, line 83: Normative interop conflict: this rule says receivers MUST enforce `data_segment_count + parity_segment_count < maxTotalSegments`, but `maxTotalSegments` is described at line 156 as 'Implementation-specific parameter, fixed. The reference implementation uses **256**'. If implementations pick different values, they will reject each other's messages even when the wire format is otherwise valid. Either pin `maxTotalSegments = 256` as a wire-format constant in this section (so it's part of the protocol, not the implementation), or relax this MUST to a SHOULD with implementation-defined caps. The DoS section at line 192 ('Validate segment counts (≤ 256)') already implies a normative cap, which suggests the former is the intent.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,223 @@
...
+  the reconstructed payload SHALL be delivered to the application.
+
+---
+
+## Implementation Suggestions
+
+### Reed–Solomon
+
+Implementations that apply parity SHALL use fixed-size shards of length `segmentSize`.
+The reference implementation uses **nim-leopard** (Leopard-RS) with a maximum of **256 total shards**.
```
Cofson, line 137: Two Reed-Solomon questions worth pinning before any second implementation lands:

1. Is RS support REQUIRED for *receivers* (so they can recover from lost data segments when the sender opts in)? Line 17 says RS is 'optional' and line 103 says senders 'MAY use' it, but a receiver that doesn't support RS will silently fail on any message with lost data segments. Suggest: 'Senders MAY use RS; receivers MUST support RS decoding to interoperate with senders that opt in.'

2. The spec names `nim-leopard` (Leopard-RS) as the reference but doesn't pin the on-wire RS variant — field (GF(2^8) vs GF(2^16)), generator polynomial, shard ordering. Two implementations that both 'use Reed-Solomon' will not interoperate without agreement on these. Pin the variant explicitly (Leopard-RS is well-defined: GF(2^16), specific encoding) or call out that the reference implementation's exact parameters are normative.

Cofson (commented):
Useful migration, the spec is compact and the wire format reads cleanly. A few interoperability gaps worth pinning down before draft:

fbarbu15 (approved):

PR #334, opened 2026-05-11
chore: split ift ts specs

fbarbu15:
## Summary

- Reorganized `ift-ts` docs by component and lifecycle status:
  - Moved most of the `ift-ts` specs to `anoncomms`
  - Added `research` section where we have COSS
  - Moved MVDS specs under Messaging
  - Split Messaging specs into `raw`, `draft`, `stable`, `deprecated`, and `deleted` folders instead of legacy folders
  - Updated Research and AnonComms README copy.

- Updated generated navigation/index support for the new component names and status folders.
- Added validator coverage so status-scoped components catch folder/status mismatches.
- Cleaned links after the moves and fixed lint issues surfaced by the restructure.
- Made history generation tolerate local renamed paths before they are tracked by git.


Cofson (approved):
This looks good - thank you! My only comment would be that 1/COSS is not technically research but it works like this for now.

PR #335, opened 2026-05-12
Add SURB specification [rebased]

AkshayaMani:
This PR adds §8.7 Single-Use Reply Blocks (SURBs) to `mix.md`, specifying how SURBs are structured, created, used by recipients, processed by mix nodes, and recovered by the Exit Layer.

Rebased onto `master` from #307 (which targeted the now-deleted `chore/mix-restructure`). All review comments are already addressed.

### DIFF
--- a/docs/anoncomms/raw/mix.md
+++ b/docs/anoncomms/raw/mix.md
@@ -660,7 +660,7 @@ Payload &= 4608 - 624 \\
 `$
 
 Implementations MUST account for payload extensions, such as SURBs, when determining the maximum message size that can be encapsulated in a single Sphinx packet.
-Details on SURBs are defined in [Section X.X].
+Details on SURBs are defined in [Section 8.7.1](#871-surb-component-sizes).
 
 The following subsection defines the padding and fragmentation requirements for ensuring this fixed-size constraint.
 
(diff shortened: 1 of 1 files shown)


docs/anoncomms/raw/mix.md
```diff
@@ -1232,9 +1240,215 @@ Once the node determines its role as an exit following the steps in [Section 8.6
...
+
+4. **Assemble SURB**
+
+   Assemble the SURB tuple $\mathrm{hop}_0, (α_0, β_0, γ_0), \tilde{k}$ as defined in [Section 8.7.1](#871-surb-component-sizes). Here $\mathrm{hop}_0$ is the address of the first hop on the return path. Embed the SURB in the message payload.
+
+   The initiating node MUST store the tuple $(\tilde{k}, s_0, \ldots, s_{L-1})$ in a local table indexed by $\mathrm{id}$, for reply recovery defined in [Section 8.7.5](#875-reply-recovery).
+
+#### 8.7.3 Using a SURB
+
+When the Exit Layer receives any SURBs as part of the handoff defined in [Section 8.6.4](#864-exit-processing) Step 4, it MUST retain them for routing the destination's responses back to the sender.
```
chaitanyaprem, line 1333: I am wondering in implementation do we have a cleanup or a timeout for how long to retain a SURB header in exit layer?  Similarly in the EntryLayer how long does the mapping of SURB-ID to keys maintained? 
Should we indicate an implementation detail that this needs to be cleaned-up somehow?

chaitanyaprem (approved):
Minor comment, otherwise looks good to me.
Thanks!

docs/anoncomms/raw/mix.md
AkshayaMani: Good catch. Looking at the implementation:
- Initiating node side: the SURB-ID → keys mapping has no TTL and only clears on receiving a matching reply, leaving a real leak when replies never arrive.
- The Exit Layer uses SURBs inline (nothing stored), so no cleanup is needed there in practice.

Added cleanup guidance to both §8.7.2 and §8.7.3.

PR #336, opened 2026-05-13
fix: segmentation RS compatibility issues

igor-sirotin:


Cofson (commented):
Thanks for the fast turnaround on the three #331 items — all three reads cleanly now. A few additional things worth pinning down before this leaves raw, with suggestions noted inline.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,237 @@
...
+A message is either a **data segment** (when `is_parity == false`) or a **parity segment** (when `is_parity == true`).
+Data segments at index `< data_segment_count - 1` carry exactly `segmentSize` bytes in `payload`. The last data segment (`data_segment_index == data_segment_count - 1`) carries between `1` and `segmentSize` bytes — these are the original (unpadded) bytes of the final chunk.
+
+### Validation
+
+Receivers MUST enforce:
+
+- `entire_message_hash.length == 32`
+- `data_segment_count >= 1`
+- `data_segment_count + parity_segment_count <= 256` (see [`maxTotalSegments`](#configuration))
```
Cofson, line 89: **Suggestion:** the Validation list captures count/index invariants but skips the payload-size invariants the wire format relies on. A receiver implementing only the listed checks would still accept a malformed segment.

Worth adding (before this leaves raw):

- non-last data segments (`data_segment_index < data_segment_count - 1`): `payload.length == segmentSize`
- last data segment (`data_segment_index == data_segment_count - 1`): `1 <= payload.length <= segmentSize` (per line 81)
- parity segments (`is_parity == true`): `payload.length == segmentSize` (per Reed–Solomon section, line 142)

These are stated as invariants elsewhere in the doc; folding them into the Validation MUSTs makes the receiver contract self-contained.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,237 @@
...
+- `original_payload_length >= 1`
```
Cofson, line 90: **Suggestion:** `original_payload_length >= 1` is the only bound currently enforced, but the field is cross-tied to `data_segment_count` and `segmentSize` — a malformed message can carry a self-inconsistent triple that this rule won't catch (e.g. `data_segment_count = 1` with `original_payload_length = 10 * segmentSize`, or `data_segment_count = 5` with `original_payload_length = segmentSize`).

Worth tightening to:

```
(data_segment_count - 1) * segmentSize < original_payload_length <= data_segment_count * segmentSize
```

i.e., every data segment except the last is full, and the last segment carries at least 1 byte. That makes the (count, length) pair a single check rather than relying on the receiver to catch the mismatch only at the truncation step.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,237 @@
...
+  On mismatch,
+  the message MUST be discarded and logged as invalid.
+- Once verified,
+  the reconstructed payload SHALL be delivered to the application.
+
+---
+
+## Reed–Solomon
+
+Implementations that apply parity MUST use fixed-size shards of length `segmentSize` and SHALL use [Leopard-RS](https://github.com/catid/leopard) as the erasure coding algorithm.
```
Cofson, line 142: **Suggestion:** `SHALL use [Leopard-RS] as the erasure coding algorithm` reads as binding implementations to one specific library, but the interop requirement is the encoding, not the library binary. A Nim/Rust/Go re-implementation that matches the same Reed–Solomon code on the wire should still interoperate.

Worth rephrasing along the lines of:

```
SHALL be wire-compatible with Leopard-RS's Reed–Solomon construction at the
specified shard size, so any implementation can interoperate with senders
that use Leopard-RS.
```

This keeps the canonical reference but doesn't preclude conformant alternative implementations.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,237 @@
...
+  REQUIRED parameter, configurable by the application. Sender and receiver implementations within the same application MUST agree on `segmentSize`.
+- `parityRate` — fraction of parity shards relative to data shards.
+    Configurable by the application. Defaults to **0.125** (12.5%). A value of `0` disables Reed–Solomon parity.
+- `maxTotalSegments` — maximum number of total shards (data + parity) per message.
+    Fixed wire-format constant: **256**.
+
+**Maximum payload size:**
+The maximum supported `original_payload_length` is bounded by `maxTotalSegments` and `segmentSize`:
+`maxPayloadBytes = floor(maxTotalSegments / (1 + parityRate)) * segmentSize`.
+With `parityRate = 0.125` and `maxTotalSegments = 256`, this gives up to **228** data segments per message (and up to **256** when `parityRate = 0`).
```
Cofson, line 157: The formula and the stated limit don't quite line up. With `parityRate = 0.125` and `maxTotalSegments = 256`:

```
floor(256 / (1 + 0.125)) = floor(227.555...) = 227
```

so the formula gives **227** data segments. The text says "up to **228** data segments per message". Two paths back to consistency:

- if the parity count is computed as `floor(data * parityRate)` (so 228 data + 28 parity = 256 fits), then the formula should be `floor((maxTotalSegments * 1.0) / 1.125)` rounded toward the floor in the same direction — but it's already `floor()`, so the formula needs an explicit `ceil()` somewhere else
- if parity is `ceil(data * parityRate)` (the stricter reading: data=228 → parity=29 → total=257 → overflow), then 227 is the correct number and the text needs updating

Either is fine, just worth aligning so implementations pick the same cap.

docs/messaging/raw/segmentation.md
```diff
@@ -0,0 +1,237 @@
...
+`entire_message_hash` enables correlation of segments that belong to the same original message but does not reveal content.
+To prevent this correlation, applications SHOULD encrypt each segment after segmentation (see [Encryption](#encryption)).
+Traffic analysis may still identify segmented flows.
+
+### Encryption
+
+This specification does not provide confidentiality.
+Applications SHOULD encrypt each segment after segmentation
+(i.e., encrypt the serialized `SegmentMessageProto` prior to transmission),
+so that `entire_message_hash` and other identifying fields are not visible to observers.
```
Cofson, line 194: **Suggestion:** the Abstract (line 19) says "All messages are wrapped in a `SegmentMessageProto`", which reads as the on-the-wire representation. Combined with this section recommending `SegmentMessageProto` itself be encrypted, the wire-format invariant becomes "all messages are wrapped in a `SegmentMessageProto`, optionally itself encrypted by the application" — worth making that layering explicit:

- the inner wire format is `SegmentMessageProto`
- the outer transport may carry `SegmentMessageProto` directly, or a ciphertext that decrypts to `SegmentMessageProto` when the application provides confidentiality
- key/algorithm coordination is the application's responsibility (out of scope)

Otherwise an implementation reading just the Abstract may not realise that the receiver of an encrypted segment first decrypts the transport payload to get the `SegmentMessageProto` before any of the Wire Format validations apply.

PR #341, opened 2026-05-20
Extend Mix Cover Traffic for stake-weighted RLN; queue forwarded packets

chaitanyaprem:
## Summary

Adapts the Mix Cover Traffic specification (`docs/anoncomms/raw/mix-cover-traffic.md`) to integrate with the Stake-Weighted Mix RLN DoS Protection introduced on the parent branch, while staying compatible with flat RLN and rate-limit-only deployments. Also resolves the two open future-work items on the stake-weighted spec that depended on this work.

## Cover-traffic spec changes

**Three explicit rate variables introduced and used throughout:**
- `R_node` — per-node rate-limit budget (may vary under stake-weighted RLN)
- `R_base` — deployment-wide anchor rate (uniform)
- `R_cover` — deployment-wide cover-emission budget = `f × R_base` (uniform)

**Redefined `cover_rate_fraction f`** such that `R_cover = f × R_base`, replacing the prior framing that conflated cover ceiling with per-node slot budget. Fixes the "30% of per-node slot budget" wording (correct as a fraction of the cover ceiling, but misleading as an absolute claim against `R_node`).

**§4 Rate Limit Budget Model rewritten** to anchor cover sizing against `R_cover` and show how the cover bound holds under both uniform and differentiated `R_node`. New "Why `R_cover` is uniform" subsection argues from forwarding-capacity under uniform random path selection (low-`R_node` forwarders cannot absorb per-node-scaled cover if high-`R_node` nodes drive the average up).

**Deployment-config guidance** for deployments without a DoS-protection mechanism: proof-related interface calls are no-ops, and `R_node`/`R_base` must still be configured uniformly across all nodes.

**Forwarded-packet behaviour changed** on `ClaimSlot` failure: queued for the next epoch instead of dropped. Per-hop generated proofs allow fresh proof generation at re-emission, so cross-epoch queueing is safe. New §9.5 specifies implementation rules — one-epoch lifetime cap, bounded queue with tail-drop, FIFO ordering, staggered flush, fresh proof, local-only diagnostics.

**Tightening pass:**
- §5.4 cover packet reception: drops redundant self-exit re-derivation; keeps codec-visibility and accumulation-attack rationale.
- §6 intro and §6.3/§6.4: defer to §5 procedures rather than restating failure handling.
- Epoch-boundary cross-arrival note: generalised to all packets (not cover-only) and root-causes mixing delay.
- Indistinguishability claims deduplicated; wire-format paragraph trimmed.
- Tier-privacy framing demoted from co-equal reason in §4 to a one-line cross-ref in §10.3.

**Reference fixes:** `mix-spam-protection-rln.md` → `mix-dos-protection-rln.md` (rename happened upstream); `mix-dos-protection-rln-stake-weighted.md` added to references.

## Stake-weighted RLN spec changes

- New §4.5 **Cover Traffic Budget** — points at the cover-traffic spec, states `R_cover = f × R_base` is uniform, and notes `(R_base, R_max]` is unused for cover by design.
- "Rate-weighted path selection" moved from §7 Future Work to §6 Out of Scope, with the rationale that weighted selection would create surveillance hotspots for high-stake nodes and break the cover-emission sizing analysis.
- "Cover traffic under stake-weighted rates" removed from §7 Future Work — resolved by this PR.


jm-clius (changes requested):
It seems like a base branch inconsistency makes this PR difficult to parse with 192 file changes. :D Happy to review the diff once it's clean.

chaitanyaprem:
> It seems like a base branch inconsistency makes this PR difficult to parse with 192 file changes. :D Happy to review the diff once it's clean.

Wow, did not realize this. will fix it and push so that you can review.

Cofson (commented):
Thanks for the careful rework — the split into `R_node` / `R_base` / `R_cover` makes the budget model legible in a way it wasn't before, and the matching §4.5 in the stake-weighted spec closes the loop nicely. The 192-file noise from the chained base branch makes the GitHub UI hard to navigate, agree with the previous reviewer there; for what it's worth, the substantive content delta is just the two files modified by commit `57d421e` (mix-cover-traffic.md ±310 lines and mix-dos-protection-rln-stake-weighted.md ±45 lines), which is reviewable in isolation. A few suggestions below on the new queueing behavior and the rationale wording.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,819 @@
...
+  The actual cover output per node therefore varies with network conditions.
+
+The slot pool is self-balancing — no explicit origination rate constraint is needed.
+Heavier forwarding load automatically leaves fewer slots for cover; lighter load leaves more.
+Under differentiated `R_node`, any per-node capacity in `(R_base, R_node]` is exclusively available to non-cover origination and forwarding;
+cover never draws from this headroom.
+
+**Why `R_cover` is uniform.**
+Under uniform random path selection, each node's forwarding load tracks the network-average origination rate, not its own `R_node`.
+Per-node `R_cover` would let high-`R_node` nodes drive the average up, and low-`R_node` forwarders would drop the cover packets they cannot absorb.
```
Cofson: **Suggestion (§4 — rationale wording):** "low-`R_node` forwarders would drop the cover packets they cannot absorb" reads as if the new behavior were to drop, but §5.2 + §9.5 in this same PR switch forwarded-packet handling from drop to queue. The argument itself is still sound (a counterfactual per-node `R_cover` would now cause queue overflow / large persistent backlogs at low-`R_node` forwarders, which is just as broken), but the wording is a leftover from the old drop-on-exhaustion semantics. Consider rephrasing to "low-`R_node` forwarders would be unable to absorb the resulting forwarding load" or similar that does not depend on the dropped/queued distinction.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,819 @@
...
+via `GenerateProof(binding_data)` ([Mix DoS Protection §8.2.1](mix-dos-protection.md#821-proof-generation)),
+where `binding_data` is the packet-specific data as defined by the DoS protection mechanism.
+
+If the claim fails, the packet SHOULD be handled as follows to avoid hitting DoS protection limits:
+
+- **Locally originated messages**: queued for the next epoch.
+- **Forwarded packets**: queued for the next epoch.
+  Since per-hop generated proofs are produced at send time ([§4](#4-rate-limit-budget-model)),
+  the queued packet is re-emitted in the next epoch with a fresh proof generated then.
+  Implementations MAY bound the queue size; packets that would exceed the bound are dropped.
```
Cofson: **Suggestion (§5.2 — per-sender accounting for queued forwards):** the bounded-queue + tail-drop policy is implementation-defined, but the queue itself is shared across all upstream forwarders at a given node. A single adversarial sender pushing high-rate forwards at epoch end could fill the queue and cause tail-drops of honest forwards arriving immediately after. Consider adding a note that implementations SHOULD partition the queue (or apply per-sender quotas) so that one upstream cannot exhaust the buffer at the expense of others. This mirrors the DoS analysis in [Mix Cover Traffic §10] / §9.5 where queue depth is already treated as a privacy-sensitive counter.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,819 @@
...
+forwarded packets that fail to claim a slot are queued for the next epoch.
+Implementations SHOULD observe the following:
+
+- **One-epoch lifetime cap**:
+  A queued packet MUST be either re-emitted within the next epoch or dropped at that epoch's end.
+  Re-queueing across multiple epochs MUST NOT be permitted.
+  This bounds the latency penalty to at most one epoch
+  and avoids cumulative synchronization effects in flush bursts.
+- **Bounded queue size**:
+  Implementations MUST configure a maximum queue size.
```
Cofson, line 636: **Suggestion (§9.5 — queue sizing guidance):** "Implementations MUST configure a maximum queue size" with no concrete guidance leaves the operational dial uncalibrated. A natural upper bound is the non-cover headroom one epoch can absorb, e.g. `R_node − R_cover/(1+L)` minus an expected non-cover origination budget, since anything larger cannot drain within the one-epoch lifetime cap. Even just stating "implementations SHOULD size the queue to what one full epoch of non-cover slots can drain" would help implementers and aligns with the lifetime-cap principle directly above.

docs/ift-ts/raw/mix-cover-traffic.md
```diff
@@ -0,0 +1,819 @@
...
+by assigning all outgoing packets — cover, locally originated, and forwarded —
+to a shared fixed-time grid determined at epoch start.
+
+### 10.5 Cover Priority and Forwarded Packet Deferral
+
+Cover emissions occur on a schedule independent of forwarding load,
+so slots consumed by cover early in an epoch are unavailable to forwarded packets arriving later.
+Under uneven forwarding load, this can cause honest forwards to be queued for the next epoch ([§6.4](#64-packet-forwarding))
+even when total traffic stays within the `R_node` budget,
+adding at least one epoch of latency.
```
Cofson, line 735: **Suggestion (§10.5 — timing privacy of the new queue):** moving forwarded packets from drop to one-epoch deferral is an unambiguous availability win, but it does introduce a new observable: queue depth at a node manifests as a per-node latency signal across epoch boundaries. A global passive adversary could potentially distinguish congested forwarders (long queues at epoch flush) from quiet ones (immediate forwards), which is an additional axis of node fingerprinting that §10.3 does not yet cover. Worth either an explicit acknowledgment here or a cross-reference from §10.3 noting that the queue introduces a second observable beyond cover-rate perturbation.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
```diff
@@ -309,6 +321,14 @@ The following are explicitly out of scope for this specification:
 - Detailed migration procedures and tooling for networks upgrading from flat-rate RLN
 - Voluntary deregistration mechanisms (a [RLN Per-Hop DoS Protection](./mix-dos-protection-rln.md) responsibility)
 - Membership registry implementation (smart contract, coordination layer, or other)
+- **Rate-weighted path selection.**
```
Cofson, line 324: **Comment (§6 — promoting rate-weighted path selection to Out of Scope):** moving this from Future Work to Out of Scope is a stronger design commitment than the previous wording, and the "surveillance hotspots" framing makes the rationale clear. One small thing: the previous Future Work entry left the door open to weighted selection as a future alternative. Stating it as out-of-scope effectively closes that door for this specs lifetime — worth confirming with the wider Mix team that this is the intended posture, since a downstream RFC would otherwise need to revisit it.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Good catch, this was a leftover. Fixed:

> Per-node `R_cover` would let high-`R_node` nodes drive the average up,
> leaving low-`R_node` forwarders unable to absorb the resulting forwarding load.

As you note, the argument itself is unaffected — the wording now doesn't depend on whether the overflow manifests as a drop or a backlog.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Agreed, and added. The rule lives in §9.5 rather than §5.2, since §5.2 is procedural and this PR already moved the other queue mechanics into §9.5:

> - **Per-predecessor fairness**:
>   The queue is shared across every upstream node a forwarder receives packets from.
>   A single upstream sending at high rate near epoch end can otherwise fill the queue
>   and cause tail-drops of packets forwarded on behalf of others.
>   Implementations SHOULD partition the queue per immediate predecessor,
>   or apply a per-predecessor quota,
>   so that tail-drop applies within a partition rather than across the whole buffer.

One thing worth stating explicitly, since §9.5 otherwise treats queue state as privacy-sensitive: per-predecessor partitioning does **not** add exposure. A node already observes its immediate predecessor — it's the incoming connection — so keying the queue on it reveals nothing further about the packet's path. I added a line to that effect so the rule doesn't read as being in tension with the §9.2 exposure restrictions.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Agreed that "MUST configure a maximum queue size" with no calibration was underspecified. I've added guidance, but split it into a ceiling and a target rather than a single number, because I think your `R_node − R_cover/(1+L)` is the right **ceiling** and the wrong target.

That expression is the *entire* non-cover slot headroom one epoch is guaranteed to offer — the quantity §5.2 already derives. A queue sized to all of it would consume every non-cover slot draining backlog, starving the next epoch's own forwards and local origination. So:

> The queue size MUST NOT exceed `R_node − f × R_base / (1 + L)`,
> the non-cover slot headroom one epoch is guaranteed to offer (see §5.2).
> A queue larger than this cannot drain within the one-epoch lifetime cap above,
> so the excess is capacity that can never be used.
> Implementations SHOULD size the queue to a fraction of that headroom rather than all of it,
> since queued packets compete for slots with forwards and locally originated messages
> arriving during the next epoch.

The "cannot drain within the lifetime cap" framing is exactly your point about aligning with the cap directly above — it just yields a hard bound rather than a recommendation.

docs/ift-ts/raw/mix-cover-traffic.md
chaitanyaprem: Agreed, this is a real new observable and it wasn't covered. Added a subsection to §10.5 plus a cross-reference from §10.3 as you suggested.

The framing I went with: deferral doesn't introduce a signal where there was none, it changes the *shape* of one. Drop-on-exhaustion was observable too, as packet loss; deferral converts that into latency. The latency version is the more legible of the two, because it persists for as long as the node stays congested, whereas a drop is a point event.

Resolution is the same as §10.3 and §10.4 — pre-scheduled emission timing (§11.3) decouples observable send times from queue state. I also noted that this is a property of emitted traffic, distinct from the queue-depth *counter*, which §9.5 already confines to memory.

docs/ift-ts/raw/mix-dos-protection-rln-stake-weighted.md
chaitanyaprem: Confirming: uniform random path selection staying is a deliberate, already-reviewed position, not something that slipped in with this PR. So it stays in Out of Scope.

That said, your read was right, and re-reading the diff I think the move did more than change status. The old Future Work entry carried a genuine argument *for* weighting — that under uniform selection, flooding low-stake nodes affects every honest path containing one, which weighted selection would blunt. Replacing it with a one-sided bullet dropped that analysis from the spec entirely, so anyone revisiting later would have to rederive it. Rewritten to state both directions, then give the reason it's out of scope anyway, and to say explicitly that a future specification MAY revisit the trade-off.

I also dropped the second justification — that weighted selection would invalidate the cover-emission sizing derivation. That derivation is in this same PR stack and takes uniform path selection as an input, so citing it here is closer to circular than to independent evidence. The surveillance-hotspot argument stands on its own.

PR #342, opened 2026-05-22
Per-hop RLN Proof Encryption

AkshayaMani:
This PR adds proof encryption for [RLN Per-Hop DoS Protection](https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/mix-spam-protection-rln.md). The per-hop RLN proof is appended outside the Sphinx packet, so its `epoch` field — an invariant across hops — is visible to passive observers and enables cross-hop packet correlation.
This PR encrypts the proof with the next hop's `X25519` public key, removing that linkability channel. Encryption reuses the Mix Protocol's existing `X25519` / `KDF` / `AES-CTR` primitives.
Sender-generated proofs are unaffected, since Sphinx already encrypts `β`.

#### `mix-dos-protection.md` (DoS protection framework)

- §4.2 Per-Hop Generated Proofs: Redefines `σ` as the encrypted proof and updates the construction steps
- §5.2 Node Responsibilities: Adds encrypt and decrypt steps to per-hop construction and preprocessing
- §6 Anonymity and Security Considerations: Adds a new proof confidentiality requirement — per-hop proofs MUST be encrypted with the verifying hop's `X25519` public key; sender-generated proofs need not be
- §8 DoS Protection Interface: Adds a `hop_pubkey` parameter to `GenerateProof`; `VerifyProof` decrypts `encoded_proof_data` before verification

#### `mix-spam-protection-rln.md` (RLN instantiation)

- Proof Encryption / Proof Decryption: Specifies encrypting the proof to the next hop and recovering it at the verifier
- Encrypted Proof Wire Format: Defines `σ = A | c` — the 32-byte ephemeral group element concatenated with the `AES-CTR` ciphertext of the serialized `RateLimitProof`
- Wire-Level Privacy: Records the cross-hop correlation motivation
- Framework alignment: Adopts [Mix DoS Protection](https://github.com/logos-co/logos-lips/blob/master/docs/anoncomms/raw/mix-dos-protection.md) terminology (`sender` → `initiating node`, `spam protection` → `DoS protection`), drops the redundant Node Roles section, and fixes stale cross-references and links


Cofson (commented):
Encrypting the per-hop proof closes the cross-hop `epoch`-correlation channel cleanly and the X25519/KDF/AES-CTR reuse keeps the cryptographic surface small. A few interop/security clarifications worth pinning down before this leaves raw, with suggestions inline.

docs/anoncomms/raw/mix-spam-protection-rln.md
```diff
@@ -128,90 +147,187 @@ This ensures that mix nodes can detect spam and trigger slashing.
...
+   \mathrm{proof\_aes\_key} = \mathrm{KDF}(\text{"proof\_aes\_key"} \mid z)\\
+   \mathrm{proof\_iv} = \mathrm{KDF}(\text{"proof\_iv"} \mid z)
+   \end{array}
+   `$
+
+4. Encrypt the serialized `RateLimitProof`:
+
+   $`
+   c = \mathrm{AES\text{-}CTR}(\mathrm{proof\_aes\_key},\ \mathrm{proof\_iv},\ \mathrm{RateLimitProof})
+   `$
```
Cofson, line 197: **Suggestion:** worth a line stating that AES-CTR provides confidentiality only, and that integrity of $\sigma$ is provided by the inner zk-SNARK over the public inputs (`merkle_root`, `epoch`, `share_x`, `share_y`, `nullifier`) — any bit flipped in `c` will surface as either a malformed proof structure or a failed zk-SNARK check at the verifier.

Without that note, a reader will reasonably wonder why there's no MAC, and an implementer may be tempted to add one (HMAC over `c`, AES-GCM in place of AES-CTR, etc.) which would change the wire format and break interop. Anchoring the integrity argument here means the wire format `\sigma = A \mid c` is the one source of truth.

docs/anoncomms/raw/mix-spam-protection-rln.md
```diff
@@ -128,90 +147,187 @@ This ensures that mix nodes can detect spam and trigger slashing.
...
 
-A mix node MUST validate a received message using the below checks, discard the message and stop further checks or processing on failure.
+In order to process messages received via mixnet,
+a mix node MUST decrypt the [RateLimitProof](#ratelimitproof), $\sigma$, attached to the sphinx packet
+and validate it (see [Mix DoS Protection spec](mix-dos-protection.md#42-per-hop-generated-proofs)).
 
-1. If the `epoch` in the received message differs from the mix node's current `epoch` by more than `max_epoch_gap`.
+#### Proof Decryption
+
+Upon receiving a sphinx packet with attached $\sigma$ containing $(A, c)$,
```
Cofson, line 256: **Suggestion:** "$\sigma$ containing $(A, c)$" leaves the split implicit. Worth one line spelling it out, since the wire format is fixed-size:

```
$A = \sigma[0..32)$ (the 32-byte ephemeral group element)
$c = \sigma[32..|σ|)$ (the AES-CTR ciphertext)
```

That removes any ambiguity for implementations and makes it explicit that a packet with `|σ| < 32` (or a `|σ|` that disagrees with the deployment's configured fixed size, per [DoS §4.2.4](mix-dos-protection.md#424-impact-on-packet-size)) MUST be rejected before any decryption attempt.

docs/anoncomms/raw/mix-spam-protection-rln.md
```diff
@@ -128,90 +147,187 @@ This ensures that mix nodes can detect spam and trigger slashing.
...
+
+$\sigma$ MUST be attached to the sphinx packet as explained in [sending and forwarding messages](#sending-and-forwarding-messages).
+$\sigma$ is the byte concatenation $A \| c$, where:
 
-The following `RateLimitProof` MUST be added to the sphinx packet as $\sigma$ as explained in [sending](#sending-and-forwarding-messages).
+- $A$ is the 32-byte ephemeral group element
+- $c$ is the AES-CTR ciphertext of the serialized `RateLimitProof`
+
+Both $A$ and $c$ are generated as described in [Proof Encryption](#proof-encryption).
+The size of $\sigma$ MUST be fixed (see [Mix DoS Protection §4.2.4](mix-dos-protection.md#424-impact-on-packet-size)).
```
Cofson, line 325: **Suggestion:** "$|σ|$ MUST be fixed" is the receiver's interop requirement, but the spec doesn't say how a sender achieves it given that the encoded `RateLimitProof` is the output of a protobuf serializer.

In practice the five field lengths are all fixed (`proof = 128`, `merkle_root = 32`, `epoch = 32`, `share_x = 32`, `share_y = 32`, `nullifier = 32`), so the serialised proto length is deterministic *if* every implementation uses the canonical proto3 encoding (fixed-length tag/length prefixes, no skipped default fields, fields emitted in tag order). Worth either:

- stating that `RateLimitProof` MUST be serialised with canonical proto3 encoding and giving the resulting fixed byte count (so |σ| = 32 + that count), or
- defining `RateLimitProof` as a flat `A || merkle_root || epoch || share_x || share_y || nullifier || proof` byte concatenation with fixed offsets, side-stepping protobuf entirely

Either way removes the implicit assumption that two correct implementations will agree on |σ| before any cross-implementation test exists.

docs/anoncomms/raw/mix-dos-protection.md
```diff
@@ -176,47 +242,63 @@ The following table provides a brief comparison between both integration approac
...
-Generate the initial proof $σ$ and append it after the Sphinx packet as described in [Section 4.2.1](#421-details).
+Generate the initial proof, encrypt it, and append the resulting $σ$ after the Sphinx packet as described in [Section 4.2.1](#421-details).
 The proof MUST NOT contain any identifying information.
 
 **[During Sphinx packet preprocessing](mix.md#861-shared-preprocessing):**
-Extract and verify the incoming proof $σ$ before any Sphinx processing as described in [Section 4.2.1](#421-details).
-If verification fails, discard the packet and apply any penalties or rate-limiting measures.
+Decrypt and verify the incoming $σ$ before any Sphinx processing as described in [Section 4.2.1](#421-details).
+If verification fails,
+discard the packet and apply any penalties or rate-limiting measures.
```
Cofson, line 273: **Suggestion:** the receiving rules cover "verification fails" but not the prior states — packets that arrive with no $\sigma$ at all, or with a $\sigma$ whose length disagrees with the deployment's fixed $|σ|$, or with malformed/undecryptable bytes.

For a deployment that has selected `PER_HOP_GENERATED`, those are all distinct failure modes from "valid $\sigma$ that fails verification", and currently each MUST/MAY decision is left to implementations.

Worth one explicit MUST near here:

> Packets arriving with a missing $\sigma$, a $\sigma$ whose length does not match the deployment-configured $|σ|$, or a $\sigma$ that fails to decrypt MUST be silently discarded (subject to the same anti-probing constraints as a verification failure).

PR #343, opened 2026-05-22
fix: extended peer record domain & payload type

SionoiS:
The code and the spec diverged this PR fixes this.

I also removed some redundant info in service disco.

### DIFF
--- a/docs/anoncomms/raw/extensible-peer-records.md
+++ b/docs/anoncomms/raw/extensible-peer-records.md
@@ -151,6 +151,8 @@ Extensible Peer Records MUST be wrapped in libp2p [signed envelope](https://gith
 before distributing them to peers.
 The corresponding `ExtensiblePeerRecord` message is serialised into the signed envelope's `payload` field.
 
+> Both domain, payload type and protobuf serialization are compatible with libp2p peer record so that both records can be used interchangeably.
+
 #### Signed Envelope Domain
 
 Extensible Peer Records MUST use `libp2p-peer-record` as domain separator string
@@ -161,8 +163,7 @@ for the envelope signature.
 Extensible Peer Records MUST use the hexadecimal `0x0301`
 as the `payload_type` value.
 
-> **_Note:_** this will make Extensible Peer Records a subtype of the "namespace" [multicodec](https://github.com/multiformats/multicodec/blob/0c6c7d75f1580af329847dbc9900859a445ed980/table.csv).
-> In future we may define a more compact multicodec type for Extensible Peer Records.
+> The payload type is the [multicodec](https://github.com/multiformats/multicodec/blob/0c6c7d75f1580af329847dbc9900859a445ed980/table.csv) `libp2p-peer-record` type
 
 ## Copyright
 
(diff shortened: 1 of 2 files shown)


docs/anoncomms/raw/logos-service-discovery.md
```diff
@@ -484,34 +484,7 @@ to avoid coupling the protocol to specific formats
...
-            data: <optional_metadata>
-        }
-    ]
-}
-```
-
-**Size constraints:**
-
-- Each `ServiceInfo.data` field SHOULD be ≤ 33 bytes
-- Total encoded XPR SHOULD be ≤ 1024 bytes
```
rlve, line 509: This is not needed anymore? https://github.com/vacp2p/nim-libp2p/issues/2487 can be closed?

SionoiS: > This is not needed anymore? [vacp2p/nim-libp2p#2487](https://github.com/vacp2p/nim-libp2p/issues/2487) can be closed?

it was redundant info already written, which is referenced in  https://github.com/logos-co/logos-lips/blob/master/docs/ift-ts/raw/extended-kad-disco.md

tinniaru3005 (approved):
LGTM :)

Cofson (commented):
Good catch on the divergence — the domain string was the bigger issue, and the move to a 2-byte multicodec payload type makes the envelope a lot leaner. One follow-up on the unchanged note around the multicodec category, suggested inline.

docs/anoncomms/raw/extensible-peer-records.md
```diff
@@ -153,13 +153,12 @@ The corresponding `ExtensiblePeerRecord` message is serialised into the signed e
...
 
-Extensible Peer Records MUST use `libp2p-routing-state` as domain separator string
+Extensible Peer Records MUST use `libp2p-peer-record` as domain separator string
 for the envelope signature.
-This is the same as for ordinary libp2p [routing records](https://github.com/libp2p/specs/blob/7740c076350b6636b868a9e4a411280eea34d335/RFC/0003-routing-records.md#signed-envelope-domain).
 
 #### Signed Envelope Payload Type
 
-Extensible Peer Records MUST use the UTF8 string `/libp2p/extensible-peer-record/`
+Extensible Peer Records MUST use the hexadecimal `0x0301`
```
Cofson, line 163: **Suggestion:** since the payload type is now `0x0301`, the unchanged note below this paragraph is slightly off. Checking the table at the linked SHA:

```
libp2p-peer-record, libp2p, 0x0301, permanent, libp2p peer record type
```

So `0x0301` lives in the **libp2p** namespace and is *exactly* the `libp2p-peer-record` codec — not a subtype of `namespace`. Two consequences worth calling out:

1. The Note should read "…a subtype of the **libp2p** multicodec" (or just drop the categorisation — saying it *is* the libp2p-peer-record codec is the simpler framing).
2. Re-using `libp2p-peer-record`'s codec is a deliberate design choice (XPR envelopes are indistinguishable from regular libp2p peer-record envelopes at the envelope level, and protobuf compatibility carries the XPR-only fields as unknown fields for vanilla libp2p consumers). If that's the intent, worth saying so explicitly — that's actually the most useful thing a reader can take from this section. If it isn't, then XPR probably wants its own multicodec entry.

The follow-up "In future we may define a more compact multicodec type" can probably go away too — `0x0301` is already a 2-byte varint, which is about as compact as multicodecs get.

fbarbu15 (approved):

PR #348, opened 2026-05-27
add spec marker version for trivial filtering

Ivansete-status:
This is meant to introduce a very simple filtering hint to determine whether messages/segments follow the reliable channel spec or not.

### DIFF
--- a/docs/messaging/raw/reliable-channel-api.md
+++ b/docs/messaging/raw/reliable-channel-api.md
@@ -34,6 +34,9 @@
     * [Scalable Data Sync (SDS)](#scalable-data-sync-sds)
     * [Rate Limit Manager](#rate-limit-manager)
     * [Encryption Hook](#encryption-hook)
+    * [Wire Format](#wire-format)
+      * [Spec Marker](#spec-marker)
+      * [Versioning](#versioning)
   * [Procedures](#procedures)
     * [Node initialization](#node-initialization)
     * [Outgoing message processing](#outgoing-message-processing)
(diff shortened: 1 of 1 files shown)


jazzz (commented):
I like the concept - but I don't understand how this relates to this specification. 

This spec outlines an API for an implementation. Events, function names, and return types.

This PR adds a wire format section, but there is no wire format specified. I'm not sure what I would do with this information.

fbarbu15 (approved):

docs/messaging/raw/reliable-channel-api.md
```diff
@@ -147,6 +150,44 @@ The Encryption Hook provides a pluggable interface for upper layers to inject en
...
+#### Spec Marker
+
+A short, fixed, unencrypted tag that identifies a message on the wire as belonging to
+this layer. The marker value is the ASCII string:
+
+```
+RELIABLE-CHANNEL-API/1
+```
+
+- The marker MUST be attached to every outgoing segment **outside** the encrypted payload,
```
igor-sirotin, line 166: I agree with https://github.com/logos-co/logos-lips/pull/348#pullrequestreview-4381646436 — this doesn't really specify the wire format 🤔 

Where should this marker be attached?

docs/messaging/raw/reliable-channel-api.md
```diff
@@ -459,6 +502,7 @@ types:
 - Segment metadata (message ID, segment index, total segments) is visible to network observers unless encrypted by the hook.
 - SDS acknowledgement messages are sent over the same content topic and are subject to the same confidentiality concerns.
 - Rate limiting compliance is required to avoid exclusion from the network by RLN-enforcing relays.
+- The [Spec Marker](#spec-marker) is unencrypted and identifies traffic as belonging to this layer to any on-path observer. It is not authenticated: an attacker can forge the marker on arbitrary payloads, so receivers MUST treat marker presence only as a coarse pre-filter and rely on decryption and SDS to detect malformed or unsolicited messages.
```
igor-sirotin, line 505: I still have the questions I asked in https://github.com/logos-messaging/logos-delivery/pull/3886#discussion_r3307092861:

> 1. What are the consequences of having this as unencrypted metadata?

docs/messaging/raw/reliable-channel-api.md
```diff
@@ -147,6 +150,44 @@ The Encryption Hook provides a pluggable interface for upper layers to inject en
...
+
+A short, fixed, unencrypted tag that identifies a message on the wire as belonging to
+this layer. The marker value is the ASCII string:
+
+```
+RELIABLE-CHANNEL-API/1
+```
+
+- The marker MUST be attached to every outgoing segment **outside** the encrypted payload,
+  so that receivers can evaluate it without performing decryption.
```
igor-sirotin, line 167: I still have the question I asked in https://github.com/logos-messaging/logos-delivery/pull/3886#discussion_r3307092861:

> 2. The goal of this field is to simplify the incoming pipeline, but:
    - This mechanism doesn't allow selecting 1 designated reliable channel. All instances of `ReliableChannel` will still have to try to decrypt a message. Find the match by trying all — it's ok IMO, at least this is what I expected for the first implementation. But then `Lip173Meta` doesn't really help much with it.
    - An attacker can spam us with random data tagged with `Lip173Meta`, forcing us to try to decrypt it. Though it's the same case without `Lip173Meta` 😄

Isn't ContentTopic enough for "coarse filtering"?

docs/messaging/raw/reliable-channel-api.md
igor-sirotin, line 505: Alhough [14/WAKU2-MESSAGE](https://lip.logos.co/messaging/stable/14/message.html#message-attributes) defines this usage of `meta` field exactly as intended

> The `meta` attribute, if present, contains an arbitrary application-specific variable-length byte array with a maximum length limit of 64 bytes. This attribute can be utilized to convey supplementary details to various [10/WAKU2](https://lip.logos.co/messaging/draft/10/waku2.html) protocols, thereby enabling customized processing based on its contents.

... I'm still concerned with tampering in-flight messages. An malicious intermediate Relay node can modify the `meta` field and the recipient won't be able to detect it.

PR #356, opened 2026-06-12
Improvement : MessageId must be unique per segment

darshankabariya:
While integrating SDS into the reliable channel (logos-co/logos-messaging/nim-delivery#3942), noticed that sending the same content twice gets dropped as a duplicate. The spec derives MessageId from content alone, so identical messages share an id and SDS deduplicates the second one. It was not clear from the spec who is responsible for handling this in the reliable channel.

This change adds senderId and a timestamp to the id derivation and states explicitly that MessageId must be unique for every segment.


fcecin (approved):
LGTM!

docs/messaging/raw/reliable-channel-api.md
```diff
@@ -118,7 +118,8 @@ See [SEGMENTATION](./segmentation.md).
 [SDS](https://lip.logos.co/anoncomms/raw/sds.html) provides end-to-end delivery guarantees using causal history tracking.
 
 - Each new segment to be sent, requires the following data:
-  - `MessageId`: a keccak-256([message](https://lip.logos.co/anoncomms/raw/sds.html#message)'s content) hex string (e.g. `4e03657aea45a94fc7d47ba826c8d667c0d1e6e33a64a036ec44f58fa12d6c45`), generated by the Reliable Channel API.
+  - `MessageId`: a keccak-256(`senderId` + timestamp + [message](https://lip.logos.co/anoncomms/raw/sds.html#message)'s content) hex string (e.g. `4e03657aea45a94fc7d47ba826c8d667c0d1e6e33a64a036ec44f58fa12d6c45`), generated by the Reliable Channel API.
```
NagyZoltanPeter, line 121: Can we detail from where timestamp is derived from? For clarity?

NagyZoltanPeter (approved):
Reasonable, maybe detail timestamp origin

docs/messaging/raw/reliable-channel-api.md
darshankabariya: Done, local Unix time in ns at wrap time.

